PR #7, opened 2022-05-23
Remove proto prefix from crate names

Tristyn:
I assume this meant ‘prototype’ from early testing
This doesn’t rename jet_proto_math or jet_proto_proc_macros
Better to do this now instead of when others depend on us


Tristyn:
Good to merge tho?

PR #19, opened 2022-05-30
Added prepare and precommit scripts.

partiallysorted:
Had to comment out some of the integration tests.


jetjake (approved):

davodesign84 (commented):
we should probably add husky as a devDependency rather than using the global install

davodesign84 (approved):

davodesign84:
Also as a side comment, I noticed that when I run `anchor test` I do get the following:
```
Error:
        /Users/davidescalzo/Desktop/jet-v2/programs/metadata/src/lib.rs:85:8
        Struct field "receiver" is unsafe, but is not documented.
        Please add a `/// CHECK:` doc comment explaining why no checks through types are necessary.
        See https://book.anchor-lang.com/chapter_3/the_accounts_struct.html#safety-checks for more information.
```

Tests pass brilliantly when doing so, worth adding?

callensm (approved):

PR #30, opened 2022-06-05
Switched to wrapped SOL.

partiallysorted:


jetjake (commented):
What's up with the tests failing?

jetjake (approved):
Nvm we know that the margin pool tests fail.  LGTM

PR #51, opened 2022-06-08
fix issues with running on 1.10.x

suscd:
- add hint for compiler to stop inlining
- remove unnecessary logs from `margin_borrow` to fit in compute limit
- remove use of unstable feature in ts lib to raise compute limits for
  `margin_borrow`


callensm (approved):

jetjake:
msg not supported or just no reason to log anymore?  That's fun on the extra data for the compute budget.

Tristyn (approved):

PR #52, opened 2022-06-08
fix clippy warnings and formatting issues

suscd:
also add format check to coverage workflow, remove checks from devnet workflow


callensm:
lmao I literally just did this too. i'll scrap my branch

.github/workflows/devnet.yaml
```diff
@@ -98,8 +98,6 @@ jobs:
       - name: Build Programs with Devnet
         run: |
           yarn install
-          cargo fmt -- --check
```
callensm, line 101: these should stay in as

```
cargo fmt --all --check
cargo clippy --all-targets -- -Dwarnings
```

for the case that this workflow is manually invoked.

callensm (changes requested):

callensm (approved):

PR #64, opened 2022-06-10
Refactor margin pool library class

davodesign84:
# Why?

### Separation of concerns
The current `marginPool.ts` file in the TS library conflates 2 separate concerns A) interacting with a margin pool and B) setup / management of a set of margin pools, the latter is unlikely to be used in an actual client application beside perhaps some power users edge cases.

This PR splits the current class in `Pool`, concerned with fetching data about the pool and interacting with it,  and `PoolManager` concerned with creation and loading of pools.

### Readability and simplicity of use
Most of the methods available in this class accept an array of arguments, which can make for some difficult to read code. E.g.
```typescript
    await maginPool_SOL.create(
      provider,
      ownerKeypair.publicKey,
      9_500,
      new BN(0),
      FEE_VAULT_SOL,
      SOL_oracle[0].publicKey,
      SOL_oracle[1].publicKey,
      POOLS[1].config
    )
```

This PR refactors most functions to accept a map of arguments so that a function like the above can be read as follows:
```typescript
    await manager.create({
      tokenMint: USDC[0],
      collateralWeight: 10_000,
      collateralMaxStaleness: new BN(0),
      feeDestination: FEE_VAULT_USDC,
      pythProduct: USDC_oracle[0].publicKey,
      pythPrice: USDC_oracle[1].publicKey,
      marginPoolConfig: POOLS[0].config
    })
```

Also as you may notice, when the `PoolManager` is initialised it requires `provider` and `programs` as a param and they are store in the class instance so they don't have to be passed around at each call. Optionally they can be overridden by adding them to the map argument, or set with setters `setProvider` `setPrograms`

### More detailed tests
Because tests. Mostly replaced `assert` with the more powerful `expect` which also provides more detailed error messages in comparison to the generic assertion error thrown by `assert`, and structure the longer tests in 2 step, `SETUP` `ACT` `TEST`, again, this should help readability.


Tristyn:
I'm a fan of using camelCase filenames and PascalCase for jsx components which is required by the react linter

Tristyn (approved):
This is great. Ready to merge

Each class has data dependencies for their getters. Building on Pool Manager, what if we made loading more tree-like? Here are some current getters, and their data dependencies 

- MarginAccount: { Record<MarginPools, Positions>, depositedValue, borrowedValue, accountBalance, leverage, etc }
  - PoolManager: { depositValue, borrowedValue, tvl }
  - Record<MarginPools, Pool>: { liquidty, debt, utilization, apr, apy, etc }

#### Proposal

- MarginWallet: { accounts: Record<number, MarginAccount>, tokens: AssociatedToken[] }
  - MarginAccount: { Record<MarginPools, Positions>, depositedValue, borrowedValue, balance, leverage, etc }
    - PoolManager: { depositValue, borrowedValue, tvl }
      - Pool: { liquidty, debt, utilization, apr, apy, etc }
    - SerumManager
      - OrderBook
    - OtherManager

Refreshing a margin account would not only fetch the account, but passing in optional data from the PoolManager or Serum would make the margin account be able to calculate getters more accurately.

### Subscriptions

A long term goal for the client lib is websocket based subscriptions for each class. For example, the MarginAccount could expose:
```ts
export class MarginAccount {
  static async load({ programs, owner, seed, poolManager?: PoolManager, serumManager?: SerumManager }): MarginAccount
  async refresh({ poolManager?: Record<MarginPools, Pool>, serumManager?: SerumManager })
  static subscribe({ programs, owner, seed, poolManager: EventEmitter<PoolManager>, serumManager?: EventEmitter<SerumManager> }): EventEmitter<MarginAccount>

  // Batteries-included convenience methods
  static async loadAll({ programs, owner, seed }): { account: MarginAccount, poolManager: PoolManager, serumManager: SerumManager }
  static async subscribeToAll({ programs, owner, seed }): EventEmitter<MarginAccount>
}
```

With this you can compose subscriptions: The margin account subscription will fire and return a new MarginAccount if the on-chain margin account is updated, or any managers or their children are updated. Load can be composed with previous loads as well.

### Immutability

Load() returns a new MarginAccount. Calling refresh() directly does not give immutability but may be convenient for lib integrators. Subscriptions return new instances of the MarginAccount making class data immutable but fresh. Lib integrators can safely build transactions against immutable data while the on-chain accounts may be mutating underneath them

davodesign84:
Yes. This. It's pretty much what I had in mind with the refactor :) Eventually expanding it to the rest of the library.

Couple of observations:

- There might not be a need to separate MarginAccount and PoolManager. In fact (provided that we can pass optional configs), the MarginAccount should be responsible for managing the pools it is interested in and can `extend` the `PoolManager` class
- A class shouldn't have a `load` method to load itself, we should have that logic inside the constructor. If some other method needs to be called to init the instance of the class the general convention is to have a method called `init` that is invoked in the constructor.
- More convention notes: Classes and the files they are defined in, should be PascalCase  to identify them at first glance but I'm ok either way.
- Immutability: I'm not sure I follow this. Initialising new instances of MarginAccount at each update sounds expensive...
- Subscriptions: Definitely one to bring in and we should extend our relevant classes from `EventEmitter` so that the flow goes something like the following:

Client instantiates the library
```typescript
// some-init-file.ts
import JetClient from '@jet-lab/margin'
import customConfig from 'yourCustomConfig.json'

const margin = new JetClient(customConfig);

export default margin;
```

Client imports the instantiated client and updates the state
```typescript
// some component
import client from './some-init-file'
import { useEffect, useState } from 'react'

const Comp = () => {
  const [price, setPrice] = useState(client.USDCPrice)
  useEffect(() => {
    client.on('update-USDC-price', (val) => { setPrice(va) })
  })
  return <div>{price}</div>
}
```
Now the component should sync with the state without having to deal with wild react tree re-renders. Or optionally you can use recoil-sync `syncEffects` to achieve the same. Init once, use everywhere, only re-render what you need to update.

PR #67, opened 2022-06-13
feature: max leverage configured per pool/token

suscd:
This replaces the c-ratio health check with one based on looking at the required collateral an account should have. This allows for configuring each token with it's own max leverage, which then is used to determine the collateral requirements for debts in that token type.

The collateral weight and new max leverage configuration are now both specified in terms of percentages instead of basis points.


Tristyn (dismissed):

eyevz (dismissed):
Thanks @suscd! The margin accounting changes look good to me now.

We should think about the max c-ratio check for allowed liquidator actions, but I'll follow that up with @dnut separately from this PR.

PR #86, opened 2022-06-20
Adding overall deposited and borrowed value, so we can calc maxTradeAmounts

loganjross:


Tristyn (changes requested):
I would keep it like this
`maxTradeAmounts: Record<TradeAction, TokenAmount>`

I'd like to review the branch before merge. I published 0.1.16 so you can keep going on the frontend

Tristyn:
The lib ended up calculating max amounts a different way so this can be closed

PR #91, opened 2022-06-23
Log repayment values

nevi-me:


eyevz (dismissed):
Smol log is accepted. It would be neat if we could deserialise the Anchor Event structs conveniently in the test environment.

programs/margin-pool/src/instructions/margin_repay.rs
```diff
@@ -127,11 +127,21 @@ pub fn margin_repay_handler(ctx: Context<MarginRepay>, max_amount: Amount) -> Re
     // First record a withdraw of the deposit to use for repaying in tokens
     let withdraw_amount =
         pool.convert_deposit_amount(Amount::tokens(max_repay_tokens), withdraw_token_rounding)?;
+    msg!(
```
dnut, line 130: Isn't logging costly? Usually I try to avoid logging anything unless there is definitely an error. For example, you could make this contingent on the `withdraw` call below returning an error.

tests/integration/pool/marginPoolConfig.test.ts
```diff
@@ -1,69 +0,0 @@
...
-  })
-
-  it("should have an available liquidity", async () => {
-    const USDC = pools.USDC
-    const availableLiquidity = await getTokenBalance(provider, undefined, USDC.addresses.vault)
-    expect(bnToNumber(USDC.depositedTokens.tokens)).to.eq(availableLiquidity)
-  })
-
-  it("should have a deposit APY", async () => {
-    expect(pools.USDC.depositApy).to.eq(0)
```
dnut, line 59: Why did you remove these tests? Was it because of a failure here?

I was oddly seeing a test failure here, even on master. I think it's supposed to actually be `to.not.eq`, like the tests below. I made that change in my branch and the tests here all passed reliably for me.

programs/margin-pool/src/instructions/margin_repay.rs
suscd: It is costly, we can keep for now to help with debugging, and remove later when it's no longer necessary.

PR #100, opened 2022-06-27
Added TS code to close positions so a user can close their account.

partiallysorted:


Tristyn (dismissed):
It's good. There's a small bit of duplicate work in #98 it'd be nice to merge that first

Tristyn (approved):

PR #102, opened 2022-06-28
Refactoring `margin-pool` Instruction Interface

zachanon:
The purpose of this PR is to refactor the instruction API to make it a bit more flexible and friendly.

### Generic Repayment
There is added an instruction, `Repay`, which allows an arbitrary token holder to repay down the balance on a margin pool position. This is more flexible than the existing `MarginRepay`, which requires the tokens used for repayment to be deposited into the margin account beforehand.

### Idempotent Instructions
Inspired by the draft PR by @Tristyn [here](https://github.com/jet-lab/jet-v2/pull/48), there is added a new field in the instruction methods, `ChangeKind`. This enum will allow the user to specify whether to shift the existing account value by the given token value or to attempt to calculate the requisite notes and set the internal note balance to reflect the given token value automatically, without an exact specification of the change.

##### Example usage
```
// deposit 100 tokens
user.deposit(&mint_key, &user_token_account, ChangeKind::ShiftBy, 100 * ONE_TOKEN)?;

// withdraw all tokens in the account
user.withdraw(&mint_key, &user_token_account, ChangeKind::SetTo, 0)?;
```


Tristyn:
After `repay` is there a 2nd instruction I could call to update the margin account? So that it'd have the exact same behavior as `margin_repay`

zachanon:
`margin_repay` only updates the loan notes token account and does not mutate the margin account https://github.com/jet-lab/jet-v2/blob/master/programs/margin-pool/src/instructions/margin_repay.rs#L31

The difference here is repaying using funds from the margin account itself, or using an outside token account.

suscd:
The pool doesn't have permission to modify the margin account, the margin program will just notice any balance changes automatically in a normal invoke (or otherwise would have to be updated separately)

programs/margin-pool/src/instructions/margin_repay_from_wallet.rs
```diff
@@ -0,0 +1,137 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use anchor_lang::prelude::*;
+use anchor_spl::token::{self, Burn, Token, TokenAccount, Transfer};
+use jet_margin::MarginAccount;
+
+use crate::{events, state::PoolAction, Amount, ChangeKind, ErrorCode, MarginPool};
+
+#[derive(Accounts)]
+pub struct MarginRepayFromWallet<'info> {
```
dnut, line 25: This should be called `Repay`

programs/margin-pool/src/instructions/margin_repay_from_wallet.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// The margin account being executed on
+    #[account(signer)]
+    pub margin_account: AccountLoader<'info, MarginAccount>,
```
dnut, line 28: This account should not be needed at all. But I think you'll need my changes from #111 for that to be possible. Anyone should be allowed to repay anyone else's loan.

dnut (changes requested):

programs/margin-pool/src/instructions/margin_repay_from_wallet.rs
dnut: #111 is merged but #138 if merged would also make this easier/cleaner to implement

Tristyn:
I'm merging master into this PR while keeping borrowing to wallet functioning

programs/margin-pool/src/instructions/register_loan.rs
```diff
@@ -30,6 +30,8 @@ pub struct RegisterLoan<'info> {
     /// This will be required for margin to register the position,
     /// so requiring it here makes it easier for clients to ensure
     /// that it will be sent.
+    ///
+    /// CHECK:
```
dnut, line 34: We don't need these any more since `skip-lint` is enabled in Anchor.toml. Upgrade your anchor cli.
```suggestion
```

zachanon: agreed, but lets fix this in a separate pr

dnut: I'm skeptical of this linting feature. It's like anchor decided that anchor is the center of the universe and it's unacceptable to rely on other tools to do anything that anchor can do. There are a lot of reasons to have unchecked accounts. When you see an AccountInfo with no checks, It's already clear that the account is not being checked by anchor and you can easily look at how it is used. Usually it's being passed into some other program or library that has its own checks, and it's not meaningful to have checks at this level.

I'm not opposed to anyone adding these comments, but I find it annoying that anchor has made it a requirement to document your code in a certain way before compiling it. Often that means we end up putting these meaningless empty comments which is even worse than skipping linting.

programs/margin-pool/src/instructions/repay.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::{events, state::PoolAction, ChangeKind, ErrorCode, MarginPool, TokenChange};
+
+#[derive(Accounts)]
+pub struct Repay<'info> {
+    /// The pool with the outstanding loan
+    #[account(
+        mut,
+        has_one = loan_note_mint,
+        constraint = margin_pool.vault == pool_vault.key()
```
suscd, line 30: use `has_one`

programs/margin-pool/src/lib.rs
```diff
@@ -53,26 +54,31 @@ mod jet_margin_pool {
     }
 
     /// Deposit tokens into the pool in exchange for notes
-    pub fn deposit(ctx: Context<Deposit>, amount: u64) -> Result<()> {
-        instructions::deposit_handler(ctx, amount)
+    pub fn deposit(ctx: Context<Deposit>, change_kind: u8, amount: u64) -> Result<()> {
```
suscd, line 57: why is `change_kind` a u8 instead of the enum?

suscd (changes requested):

dnut (approved):
Nice work.

This PR shined a light on some issues with the design of the accounting logic. Let's revisit this later to make it easier add features without introducing bugs. Particularly we should think of ways to avoid:
- conflating deposit notes with loan notes (the type system can help us here)
- letting MarginPool balance fields get inconsistent with reality (there should probably be explicit comparisons or assignment directly from the token accounts)
- using PoolAction in the wrong context

programs/margin-pool/src/lib.rs
zachanon: Issues with getting the typescript to properly serialize the enum. Judged it to be better to get the change out by making the interface types simpler than continue arguing with the typescript.

zachanon:
> Nice work.
> 
> This PR shined a light on some issues with the design of the accounting logic. Let's revisit this later to make it easier add features without introducing bugs. Particularly we should think of ways to avoid:
> 
> * conflating deposit notes with loan notes (the type system can help us here)
> * letting MarginPool balance fields get inconsistent with reality (there should probably be explicit comparisons or assignment directly from the token accounts)
> * using PoolAction in the wrong context

Agree strongly. The `Amount` conversions and `PoolAction` rounding could probably be solved via some smart abstractions to the way the `convert_amount` function works, possibly taking another sort of context enum like the `PoolAction` signifying why the conversion is being done, which will inform the rounding and conversion rate.

suscd (approved):

PR #105, opened 2022-06-29
Flight logs

davodesign84:
Adding `getFlightLogs` to `MarginClient`.

Use as follows:
`const transactions = await MarginClient.getFlightLogs(provider, publicKey, mints, "localnet")`

`Mints` should have the following type
```typescript
interface TokenMintsList {
  tokenMint: PublicKey
  depositNoteMint: PublicKey
  loanNoteMint: PublicKey
}
type Mints = Record<string, TokenMintsList>
```

The aim later on would be to instantiate a MarginClient object from the MarginClient class which would do a bunch of the setup work needed for the various apis, removing the need to construct the above `mints` parameter.

The returned transactions will be an array of `TransactionLog`:
```typescript
interface TransactionLog {
  blockDate: string
  time: string
  signature: string
  sigIndex: number //signature index that we used to find this transaction
  tradeAction: string
  tradeAmount: TokenAmount
  tokenAbbrev: string
  tokenDecimals: number
}
```


Tristyn (approved):
I like how modular it is

davodesign84:
> I like how modular it is

Thanks mate :)

PR #108, opened 2022-06-29
Small compute optimization

Tristyn:


suscd:
what's the delta? how small is it?

Tristyn:
@suscd @nevi-me It appears to be a major improvement

`margin_refresh_position` simulated in `npm run test`
**9_203** less compute in the fast case: 35456 -> 26253
**10_969** less compute in the slow case: 38106 -> 27137
The slow case appeared in the 3rd of 4 invocations

### Further optimization with shifting
Div and mul are the biggest gas hogs. If Number128::ONE is a power of 2 you could shift instead. I suspect you could cut gas in half or more with the below:

```rs
// from this
const ONE: i128 = 10_000_000_000
fn mul(self, rhs: Number128) -> Self::Output {
    Self(self.0.checked_mul(rhs.0).unwrap().div(ONE))
}
fn div(self, rhs: Number128) -> Self::Output {
    Self(self.0.mul(ONE).div(rhs.0))
}

// to this
const ONE: i128 = 2<<33 // 8_589_934_592
fn mul(self, rhs: Number128) -> Self::Output {
    Self(self.0.checked_mul(rhs.0).unwrap().shl(33))
}
fn div(self, rhs: Number128) -> Self::Output {
    Self(self.0.shr(33).div(rhs.0))
}
```

suscd (approved):

PR #116, opened 2022-07-01
Register position with fewer transactions

rhubarbgarden:
Within `deposit` function, create a new `withGetOrCreatePosition` function that
- takes an instruction array
- appends a `withRegisterPosition` instruction to the array


libraries/ts/src/margin/marginAccount.ts
```diff
@@ -682,7 +687,7 @@ export class MarginAccount {
   ///
   /// Returns the instruction, and the address of the token account to be
   /// created for the position.
-  async withRegisterPosition(instructions: TransactionInstruction[], tokenMint: Address): Promise<PublicKey> {
+  async withRegisterPosition(instructions: TransactionInstruction[], tokenMint: Address): Promise<TransactionInstruction[]> {
```
Tristyn, line 690: You don't need to return anything because it's handled in `instructions.push(ix)` at the end of the method
```suggestion
  async withRegisterPosition(instructions: TransactionInstruction[], tokenMint: Address): Promise<void> {
```

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -701,7 +706,7 @@ export class MarginAccount {
       })
       .instruction()
     instructions.push(ix)
-    return tokenAccount
+    return instructions
```
Tristyn, line 709:
```suggestion
```

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -623,6 +623,11 @@ export class MarginAccount {
     throw new Error("Unable to register position.")
   }
 
+  async withGetOrCreatePosition(tokenMint: Address, ixArray: TransactionInstruction[]) {
+    assert(this.info)
+    await this.withRegisterPosition(ixArray, tokenMint)
+  }
```
Tristyn, line 629: This method should have the behaviour as the orignal `getOrCreatePosition`. It
- loops and tries to find the position in `this.positions`
- calls `withCreatePosition` if it finds nothing.
- Don't return any position here. We just want to ensure it will exist
You could rename it to `withRegisterPositionIfNotExists` maybe, cause that's more like what it's doing

Tristyn (approved):

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -613,14 +613,22 @@ export class MarginAccount {
     await this.registerPosition(tokenMintAddress)
     await this.refresh()
 
+    throw new Error("Unable to register position.")
```
Tristyn, line 619: This function is called `getOrCreatePosition`, it does create the position but throws anyway

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -613,14 +613,22 @@ export class MarginAccount {
...
       const position = this.info.positions.positions[i]
       if (position.token.equals(tokenMintAddress)) {
-        return position
+        return
       }
     }
 
-    throw new Error("Unable to register position.")
+    await this.withRegisterPosition(ixArray, tokenMint)
+    await this.refresh()
```
Tristyn, line 634: There's no point to refreshing because the account only changes after the instruction is executed
```suggestion
    await this.withRegisterPosition(ixArray, tokenMint)
```

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -237,19 +237,19 @@ export class Pool {
   /// `amount` - The amount of tokens to be deposited
   async deposit({ marginAccount, source, amount }: { marginAccount: MarginAccount; source: Address; amount: number }) {
     await marginAccount.refresh()
-    const position = await marginAccount.getOrCreatePosition(this.addresses.depositNoteMint)
-    assert(position)
-
     const instructions: TransactionInstruction[] = []
+    const positionMint = this.addresses.depositNoteMint
+    await marginAccount.getOrCreatePosition(positionMint)
```
Tristyn:
```suggestion
    await marginAccount.withGetOrCreatePosition(positionMint, instructions)
```

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -237,19 +238,19 @@ export class Pool {
...
+    const positionMint = this.addresses.depositNoteMint
+    await marginAccount.withRegisterPositionIfNotExists(positionMint, instructions)
+    assert(positionMint)
 
     await this.withDeposit({
       instructions,
       depositor: marginAccount.address,
       source,
-      destination: position.address,
+      destination: positionMint,
```
Tristyn, line 250:
```suggestion
      destination: findDerivedAccount(this.programs.config.marginProgramId, marginAccount.address, positionMint),
```

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -237,19 +238,19 @@ export class Pool {
   /// `amount` - The amount of tokens to be deposited
   async deposit({ marginAccount, source, amount }: { marginAccount: MarginAccount; source: Address; amount: number }) {
     await marginAccount.refresh()
-    const position = await marginAccount.getOrCreatePosition(this.addresses.depositNoteMint)
-    assert(position)
-
     const instructions: TransactionInstruction[] = []
+    const positionMint = this.addresses.depositNoteMint
```
Tristyn, line 242: You will want to specifically use positionAddress for the call to `withDeposit` and `withUpdatePositionBalance`
```
    const positionAddress = findDerivedAccount(this.programs.config.marginProgramId, marginAccount.address, positionMint)
```

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -335,14 +336,15 @@ export class Pool {
   }
 
   async marginBorrow({ marginAccount, pools, amount }: { marginAccount: MarginAccount; pools: Pool[]; amount: BN }) {
+    const instructions: TransactionInstruction[] = []
     await marginAccount.refresh()
-    const depositPosition = await marginAccount.getOrCreatePosition(this.addresses.depositNoteMint)
+    const positionMint = this.addresses.depositNoteMint
+    const depositPosition = await marginAccount.withRegisterPositionIfNotExists(positionMint, instructions)
```
Tristyn, line 342: `depositPosition` and `loanPosition` are undefined because `withRegisterPositionIfNotExists` doesn't return anything. maybe it can derived the position address and return it?

Tristyn (changes requested):
It's really coming together :)

Often the position mint is being passed into a instruction that needs the position token address, that would error if it executes

Could you update `marginWithdraw` as well? I didn't see it on the PR

After doing another pass and making sure it looks correct, you could try packing this and testing the transactions in the frontend? :)

libraries/ts/src/margin/marginAccount.ts
rhubarbgarden: The current implementation of `getOrCreatePosition` returns an account position. This function is also only used in one implementation in `pool.ts`.

Throwing an error is necessary to prevent a type error [here](https://github.com/jet-lab/jet-v2/blob/1737f8988b9def55218f05a96c0f301f95e5316f/libraries/ts/src/margin/pool/pool.ts#L537)

libraries/ts/src/margin/pool/pool.ts
rhubarbgarden: This should already be changed to `withRegisterPositionIfNotExists`

PR #117, opened 2022-07-01
Close margin account.

partiallysorted:


tests/declarative/replicant.ts
```diff
@@ -303,9 +307,12 @@ export class Replicant {
         }
       }
     }
-    await marginAccount.refresh()
 
-    for (const position of marginAccount.getPositions().reverse()) {
+    if (dirty) {
+      await marginAccount.refresh()
+    }
```
Tristyn, line 313: This pattern of `execute -> fetch > execute` doesn't play well with browsers or hardware wallets that promp the user to sign. I suppose that's fine in this context because this is more of a tool. It's also harder to engineer around `execute -> fetch -> execute`

Tristyn (approved):
Happy tool 😄

PR #118, opened 2022-07-03
margin-pool: refactor control instructions / auto derive fee destination

suscd:
* fee destination is no longer configurable, the accounts are now created
  and owned by the control program.
* `control::register_token` renamed to `control::create_margin_pool`
* `control::configure_token` renamed to `control::configure_margin_pool`
* refactored control API in rust sdk to match other instructions with an
  `ix_builder`


programs/control/src/instructions/create_margin_pool.rs
```diff
@@ -66,14 +67,26 @@ pub struct RegisterToken<'info> {
...
+              seeds = [
+                    crate::seeds::FEE_DESTINATION,
+                    margin_pool.key().as_ref()
+              ],
+              bump,
+              space = TokenAccount::LEN,
+              payer = requester,
+              owner = Token::id()
+    )]
+    fee_destination: AccountInfo<'info>,
```
dnut, line 80:
```suggestion
    #[account(init,
              seeds = [
                    crate::seeds::FEE_DESTINATION,
                    margin_pool.key().as_ref()
              ],
              bump,
              payer = requester,
              token::mint = deposit_note_mint,
              token::authority = authority
    )]
    fee_destination: Account<'info, TokenAccount>,
```

dnut: actually i would probably do this in the pools program. why does the control program need to worry about this?

suscd: Why should the pools program? The issue is that the pool program is only responsible for collecting fees, not managing what happens to them afterwards. The pool program only recognizes a single authority: the control program, so either way you have to invoke the control program to manage any fee account, whether the pool program owns it or the control program does. So then it seems simplest to leave it with the control program, until other needs for it arise.

suscd: Anchor's initialization code runs too early, the mints don't exist until after the pool has been initialized, which occurs after anchor initialization.

dnut (approved):

PR #125, opened 2022-07-06
updating tx history

loganjross:


libraries/ts/src/margin/marginClient.ts
```diff
@@ -119,7 +123,7 @@ export class MarginClient {
     const instructions = ["repay", "borrow", "deposit", "withdraw"]
```
Tristyn, line 123: These need to be more specific cause you can pick up all sorts of false positives in the logs
```suggestion
    const instructionLogHeaders = [
      "> Program log: Instruction: MarginRepay", 
      "> Program log: Instruction: MarginBorrow", 
      "> Program log: Instruction: Deposit", 
      "> Program log: Instruction: Withdraw"
    ]
```

libraries/ts/src/margin/marginClient.ts
```diff
@@ -119,7 +123,7 @@ export class MarginClient {
...
     let tradeAction = ""
     for (let i = 0; i < instructions.length; i++) {
-      if (transaction.meta?.logMessages?.some(log => log.toLowerCase().includes(instructions[i]))) {
+      if (transaction.meta?.logMessages?.some(tx => tx.toLowerCase().includes(instructions[i]))) {
         tradeAction = instructions[i]
```
Tristyn, line 127: I think removing lowercase is possble and it's good to be as specific as possible
```suggestion
      if (transaction.meta?.logMessages?.some(logLine => logLine.includes(instructionLogHeaders[i]))) {
        tradeAction = instructionLogHeaders[i]
```

libraries/ts/src/margin/marginClient.ts
```diff
@@ -148,7 +151,7 @@ export class MarginClient {
           ) {
             token = config.tokens[tokenAbbrev] as MarginTokenConfig
             if (
-              token.symbol === "SOL" &&
+              token.mint === NATIVE_MINT &&
```
Tristyn, line 154: Can you please review the code style doc. 
https://jetprotocol.atlassian.net/wiki/spaces/JC/pages/32801235/Typescript+Codestyle

Please test your code in a frontend or unit tests before asking for a review

It's not optional. This code won't work because `PublicKey === PublicKey` and `BN === BN` don't do anything

```suggestion
              token.mint.equals(NATIVE_MINT) &&
```

Tristyn (changes requested):
Can you please review the code style doc. Specifically the stuff about equals
https://jetprotocol.atlassian.net/wiki/spaces/JC/pages/32801235/Typescript+Codestyle

Please test your changes in a frontend or unit tests before requesting a review because you might have found the `PublicKey === PublicKey` bug

Pls fix the failing test. `tsc` and `npm run test` from the root folder

libraries/ts/src/margin/marginClient.ts
loganjross: token.mint is an address not a PublicKey so i'll have to compare .toString()

libraries/ts/src/margin/marginClient.ts
loganjross: This would change in the future for pools vs other programs (when we have trading and stuff too right)

Tristyn: Other programs would have different instruction names yes

libraries/ts/src/margin/marginClient.ts
Tristyn: In that case it can be converted to public key quickly
`translateAddress(token.mint).equals(NATIVE_MINT)`

Tristyn (approved):

PR #134, opened 2022-07-08
Use `with` type functions that return ix over tx

rhubarbgarden:
This PR minimises the number of transactions that are being called in the process of creating a position by adding the relevant instructions to existing tx calls.

- Create `withGetOrCreateDepositNotePosition` function
- Prefer `withCreateAccount` instead of `createAccount`


libraries/ts/src/margin/marginAccount.ts
```diff
@@ -659,7 +659,11 @@ export class MarginAccount {
       seed = await this.getUnusedAccountSeed({ programs, provider, owner })
     }
     const marginAccount = new MarginAccount(programs, provider, owner, seed, pools, walletTokens)
-    await marginAccount.createAccount()
+    const instructions: TransactionInstruction[] = []
+    await marginAccount.withCreateAccount(instructions)
+    if (instructions.length > 0) {
+      return await provider.sendAndConfirm(new Transaction().add(...instructions))
+    }
```
Tristyn, line 666: This does the exact same thing as `marginAccount.createAccount()` so you can leave it as is
```suggestion
    await marginAccount.createAccount()
```

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -521,13 +521,24 @@ export class Pool {
...
 
+  async withGetOrCreateDepositNotePosition(
+    instructions: TransactionInstruction[],
+    marginAccount: MarginAccount
+  ): Promise<Address> {
+    const account = marginAccount.getPosition(this.addresses.depositNoteMint)
+    if (account) {
+      return account.address
+    }
+    return await this.withGetOrCreateDepositNotePosition(instructions, marginAccount)
```
Tristyn, line 539:
```
    return await this.withGetOrCreateDepositNotePosition(instructions, marginAccount)
```
This code calls itself recursively. You're missing a function called `withCreateDepositNotePosition`

Can you make `withCreateDepositNotePosition` share code with `withCreateLoanPosition`?

Tristyn:
You'll need to call `marginAccount.withCreateAccount` from `pool.deposit`

Tristyn:
Setting the merge target to borrow-to-wallet because you merged from that and it gives us a better idea of what the PR diff will look like

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -725,12 +742,22 @@ export class Pool {
     amount: PoolAmount
     destination?: TokenAddress
   }) {
-    // FIXME: can source be calculated in withdraw?
-    const { address: source } = await marginAccount.getOrCreatePosition(this.addresses.depositNoteMint)
+    // FIXME: can be getPosition
+    const source = await marginAccount.getOrCreatePosition(this.addresses.depositNoteMint)
```
Tristyn, line 746: This could be using withGetOrCreateDepositNotePosition

When you create a function like that, you should also switch all relevant code to use it

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -725,12 +742,22 @@ export class Pool {
...
+    const postInstructions: TransactionInstruction[] = []
+
+    const marginWithdrawDestination =
+      destination ??
+      (await AssociatedToken.withCreateOrUnwrapIfNativeMint(
+        preInstructions,
+        postInstructions,
+        marginAccount.provider,
+        this.tokenMint
+      ))
```
Tristyn, line 760:
```suggestion
```

This also doesn't need to be done anymore, looks like a bad merge

PR #138, opened 2022-07-08
remove CompactAccountMeta + let margin account be anywhere in instruction

dnut:
The adapter invoke parameter `account_metas: Vec<CompactAccountMeta>` is redundant. The client already specifies the same information by including the `AccountMeta`s they want to pass to the adapter as the `remaining_accounts`. The only thing that would differ is the `is_signer` field for the margin account, which the margin program can figure out without any help from the client. So I removed the `account_metas` parameter from the invoke instructions, which saves 2 bytes for every account.

In the future, maybe there will be a situation that the client doesn't want every account to go to the adapter because some are for margin. If we need that, we can satisfy the requirement by specifying a start or end index with a u8. 

I also got rid of the requirement that the margin account must always be the first account for any instruction invoked by margin. Now it can be anywhere, or it doesn't even need to be present. Previously, the assumption that the margin account comes first saved 3 bytes per instruction. With CompactAccountMeta removed, it would only save 1 byte per instruction. It doesn't seem like a meaningful enough improvement to justify the arbitrary limitation.

This enables much more flexible adapter integration because it removes the last remaining margin-specific requirement that adapters still need to meet simply to be invoked by margin. This helps us in two scenarios:
1. The adapters we write can be more general purpose. For example, margin-pools can easily let anyone repay a loan either directly from their wallet or from their margin account, without requiring redundant or ignored accounts to be included in the repayment instruction. See my comment in @zachanon's pr: https://github.com/jet-lab/jet-v2/pull/102#discussion_r915281977
2. We can more easily let other programs we didn't develop act as their own margin adapter without necessarily requiring us to write an intermediary program to act as an adapter. 

## Compatibility
This breaks API compatibility but not storage compatibility. So nothing about existing accounts should have any issues. I implemented the change in the rust and ts libraries. An upgraded backend requires an updated frontend so they have the same communication protocol.


partiallysorted (approved):

Tristyn:
Are accounts still kept out of instruction data? i.e. can solanas transaction serializer dedupe accounts?

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -926,31 +921,63 @@ export class MarginAccount {
...
       .instruction()
     instructions.push(ix)
   }
 
+  // prepares arguments for adapterInvoke, accountInvoke, or liquidatorInvoke
+  invoke({
+    adapterInstruction,
+    marginIndex, // location that marginAccount already exists in the instruction
+    offset // number of accounts required for margin program
+  }: {
```
Tristyn, line 947: Can these arguments be removed and determined from the instruction? You can definetly search `adapterInstruction` to find the margin index, not sure about the other one

Or maybe I misunderstand something

Tristyn (commented):
Just wondering if invoke in the lib can be simplified

libraries/ts/src/margin/marginAccount.ts
dnut: Yeah that's exactly how I did it in the rust library which I reference in the comment. I guess I was getting lazy at the end and just wanted to finish something that works. I'll take a look at applying the same logic here.

dnut:
> Are accounts still kept out of instruction data?

There are still no pubkeys in the instruction arguments, if that's what you mean.

> i.e. can solanas transaction serializer still dedupe accounts?

I'm not aware of this serializer. Previously, we already had implemented our own basic form of deduplication. We required the margin account to be the first account in the adapter ix so the client wouldn't have to specify the margin account twice in the list of accounts (once for margin and once for the adapter). 

What I did was make our dedupe logic a little smarter. It finds the margin account no matter where it is in the adapter ix, rather than assuming it's always first. 

But if Solana already takes care of deduplication then maybe that was pointless from the beginning? Then I can simplify this pr to just include a duplicate margin account.

libraries/ts/src/margin/marginAccount.ts
dnut: > It's actually pretty annoying to find the margin index in the lib because anchor accepts accounts in an unordered map, the dev has to read the program source or IDL
> 
> Or maybe I misunderstand something

Actually maybe you did. This is the index of the margin account in the *margin* instruction, not the adapter instruction. So these numbers only need to be provided in adapterInvoke and accountingInvoke. I already added them there, so no one will have to worry about them again, unless one of those two instructions change. Every adapter invoke will reuse the same values.

dnut: i changed it so this is no longer relevant

callensm (approved):

PR #140, opened 2022-07-08
log position when valuation function has error

dnut:
@Tristyn said it's hell to debug these errors so I put in a log message in the only place that it actually returns an error. Can't do it for collateral without sacrificing performance since they don't fail the instruction


partiallysorted (approved):

Tristyn:
This helps a bit. the position index and mint are out of reach. But it lets you search for the position address client side. Great thanks :)

Tristyn (dismissed):

dnut:
> @dnut I don't have the permission to merge bypassing codecov

Actually it's not a bad idea to test that stale claims result in an error so I added a test for it.

PR #144, opened 2022-07-11
Adds functions for projecting rates after actions

eyevz:


eyevz:
@loganjross and @Tristyn these new functions should facilitate forecasting of deposit and borrow rates after pool actions.

Tristyn (approved):

PR #152, opened 2022-07-12
upgrade anchor

dnut:
- get anchor to latest 0.25
- which also pushes solana to 1.10
- couple other tweaks needed to make cargo update happy (tokio and bytes)
- some rust code fixes were also needed for breaking changes in the libraries

### DIFF
--- a/.github/workflows/devnet.yaml
+++ b/.github/workflows/devnet.yaml
@@ -10,8 +10,8 @@ on:
   workflow_dispatch: {}
 
 env:
-  ANCHOR_CLI_VERSION: 0.25.0
-  SOLANA_CLI_VERSION: 1.10.24
+  ANCHOR_CLI_VERSION: 0.24.2
+  SOLANA_CLI_VERSION: 1.10.29
   VALIDATOR_IMG: jet-v2-test-validator
 
 defaults:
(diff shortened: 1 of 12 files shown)


.github/workflows/devnet.yaml
```diff
@@ -10,7 +10,7 @@ on:
   workflow_dispatch: {}
 
 env:
-  ANCHOR_CLI_VERSION: 0.24.2
+  ANCHOR_CLI_VERSION: 0.25.0
```
suscd, line 13: the latest published version of anchor is `0.24.2`, so this doesn't work and breaks the integration test workflow which is falsely passing even though no tests are being run.

suscd (changes requested):

.github/workflows/devnet.yaml
dnut: oops i assumed 0.25 was fully released since it was released on github

dnut: i see 0.25.0 here https://crates.io/crates/anchor-cli

suscd: I also see there's a `0.25` version for the npm package: https://www.npmjs.com/package/@project-serum/anchor

seems like the problem is just running the latest anchor cli

suscd: It's probably fine to leave the cli version at 0.24.2 for now, while the actual linked library is 0.25.

suscd (approved):

PR #155, opened 2022-07-13
Refactor Internal `Amount` token and note conversions, and rounding implementation

zachanon:
### `Amount` refactor and conversions
The current implementation of the `Amount` type allows one to mistake loan notes and deposit notes when making conversions, as the `AmountKind` enum fails to distinguish between the two.

This PR replaces the `Amount` type with a `PartialAmount` instantiated with specific information of the conversion to be done to a given `u64` token or notes value in order to create a `FullAmount`

##### Example Usage
```
let deposit_amount = pool.convert_amount(
   PartialAmount::tokens_to_deposit_notes(token_value),
   RoundingDirection::notes_emission(PoolAction::Deposit),
)?;
```
### Rounding Implementation
The `RoundingDirection::direction(...)` method proved to be a leaky abstraction, as it was developed specifically for particular `PoolAction` workflows.

This PR removes the `direction(...)` method and replaces it with several more concrete methods for specific operations.


programs/margin-pool/src/lib.rs
```diff
@@ -140,34 +136,82 @@ pub enum ChangeKind {
...
+    /// Convenience function for unwrapping the notes value
+    fn notes(&self) -> Result<u64> {
+        self.notes
+            .ok_or_else(|| error!(ErrorCode::NotesNotCalculated))
+    }
+
+    /// Convenience function for unwrapping the tokens value
+    fn tokens(&self) -> Result<u64> {
+        self.tokens
+            .ok_or_else(|| error!(ErrorCode::TokensNotCalculated))
```
dnut, line 214: This seems like it may result in runtime errors that a different design could catch at compile time.

When implementing a caller, I might assume that I can expect to always find tokens here unless the state is genuinely corrupted. But it really depends. The Option can be None without necessarily being in a *bad* state, it's just in a state intended for a specific use, while the caller doesn't realize that it's only supposed to be used in a specific situation.

A data type should have a clear meaning that doesn't require users of the type to look at how the data was instantiated. Either it definitely has some data in a specific form, or there is some ambiguity, so you should be equipped to handle every possibility. If you want some context to assume that the data is there, then you should simply pass a FullAmount or u64 to that context. If you want to optionally provide one field or another, the type should enable users of the type to ensure at compile time that they are handling every scenario. One approach would be with an enum, for example. Since this data is always instantiated with at least one value, there are three possible forms:

```rust
EitherAmount {
    Full(FullAmount),
    Notes(u64),
    Tokens(u64),
}
```
This communicates clearly to callers of the data that the data will always be there in some form, and a conversion may be needed to use the data properly. That way, they won't neglect a situation they are capable of handling, and simply propagate errors through the call stack at runtime when they make the wrong assumption. You could also implement these methods with a parameter for the exchange rate to abstract away the conversion process.

programs/margin-pool/src/lib.rs
```diff
@@ -140,34 +136,82 @@ pub enum ChangeKind {
 
 #[derive(AnchorSerialize, AnchorDeserialize, Debug, Clone, Copy)]
 pub enum AmountKind {
-    Tokens,
-    Notes,
+    DepositNotes,
```
dnut, line 139: Another approach you might consider are wrappers for note amounts. Just pass around notes as those wrapper types rather than u64. That way it will be easier to propagate clarity about note units throughout the entire codebase rather than sometimes seeing it represented as just "notes" or just some number. And you don't need to know that you have to search for some other field called amount_kind in order to interpret what "notes" means.

You could do an enum:
```rust
enum Notes {
    Deposit(u64),
    Loan(u64),
}
```
but this has the disadvantage of requiring every function that accepts notes to be able to handle either loan or deposit notes. You may want some function to require that it receives Loan notes. You could do something like this:
```rust
struct LoanNotes(u64);
struct DepositNotes(u64);
```
Then you could use those types as the fields for the enum variants above. 
```rust
enum Notes {
    Deposit(DepositNotes),
    Loan(LoanNotes),
}
```
Or you could have them implement a marker trait called Notes.
```rust
trait Notes;
impl Notes for DepositNotes {}
impl Notes for LoanNotes {}
```

Wrapping and unwrapping comes with some boilerplate so it's not always worthwhile. Just some food for thought.

programs/margin-pool/src/lib.rs
zachanon: The `Amount` type really represents a `PartialAmount`, so renaming will provide a bit of clarity. I like the conceptual separation between a partially known `PartialAmount` and a fully calculated `FullAmount`, both types exist at very different parts of the instruction process. Adding note kind information to the `FullAmount` is a bit superfluous, as it would mean unwrapping or checking enums after the conversion itself, and the pool methods would need to return the `EitherAmount` instead of a struct containing the tokens and notes to update the accounting with.

You do however make a good point that instantiating the correct `PartialAmount` is not very clear. Perhaps the instantiation methods should make the caller clarify explicitly the transformation that they are going to make on the given value. For example,
```
impl PartialAmount {
   pub fn tokens_to_deposit_notes(tokens: u64) -> Self {
      Self {
         kind: AmountKind::DepositNotes,
         tokens: Some(tokens),
         notes: None,
      }
```
Doing things this way does not have the advantage you mention of compile time checks, but allows us to keep the current type structure we have for the pool conversion methods.

programs/margin-pool/src/lib.rs
zachanon: I like the idea of wrapping notes in an enum, but currently the `FullAmount` is used in a way in which the note kind is clear.

For example, https://github.com/jet-lab/jet-v2/blob/master/programs/margin-pool/src/instructions/margin_borrow.rs#L124
adding a check or enum unwrap here is unnecessary information, as any method that will return to you a `FullAmount` would require specifying the notes needed via the `PartialAmount` instantiation method.

zachanon:
Stale

PR #156, opened 2022-07-13
limit positions to 24 unless registered by liquidator

dnut:


programs/margin/src/state.rs
```diff
@@ -180,6 +185,9 @@ impl MarginAccount {
         max_staleness: u64,
         approvals: &[Approver],
     ) -> AnchorResult<AccountPositionKey> {
+        if self.is_liquidating() && self.position_list().length >= MAX_USER_POSITIONS {
```
suscd: this looks wrong/insufficient, because it also prevents the liquidator from registering any positions past the limit. Perhaps instead `register_position` should be disabled for users while the account is being liquidated, and the check here should be `if !self.is_liquidating() && self.position_list().length >= MAX_USER_POSITIONS`

suscd (changes requested):

programs/margin/src/state.rs
suscd: Doing it that way might be a bit confusing in terms of the error returned, it's probably better to directly return an error that the account is being liquidated.

dnut: reimplemented as verify_authority with distinct error messages depending on the situation

suscd (approved):

PR #159, opened 2022-07-14
Get flight log action from tx logMessage

rhubarbgarden:
This PR changes the method in which transaction trade actions are set in flight logs.

For each transaction, the `getTransactionData` function now:
- Iterates through each transaction `logMessage` string
- Checks each string, returns true if a `tradeInstruction` is found and sets `tradeAction` to the relevant action
- Breaks after finding the first instance of a relevant `logMessage`


libraries/ts/src/margin/marginClient.ts
```diff
@@ -128,9 +129,22 @@ export class MarginClient {
       repay: "Instruction: MarginRepay"
```
Tristyn, line 129: There is "Repay" (repay from wallet) abd "MarginRepay" (repay from margin account deposit)
Both are distinct actions

Tristyn (approved):

PR #161, opened 2022-07-15
Wrap `marginBorrow` instructions in one tx only

rhubarbgarden:
- `Pool.marginBorrow` instructions are added to part of an transaction instead of being called as separate txns, minimising the number of transactions that are called and thus the number of transactions that need to be approved by the end user.

### DIFF
--- a/libraries/ts/src/margin/pool/pool.ts
+++ b/libraries/ts/src/margin/pool/pool.ts
@@ -443,10 +443,8 @@ export class Pool {
     assert(marginAccount)
     assert(change)
 
-    await marginAccount.createAccount()
-    await sleep(2000)
-    await marginAccount.refresh()
     const instructions: TransactionInstruction[] = []
+    await marginAccount.withCreateAccount(instructions)
     const position = await marginAccount.withGetOrCreatePosition({
       positionTokenMint: this.addresses.depositNoteMint,
       instructions
--- a/libraries/ts/src/margin/pool/pool.ts
+++ b/libraries/ts/src/margin/pool/pool.ts
@@ -443,12 +443,12 @@ export class Pool {
     assert(marginAccount)
     assert(change)
 
+    await marginAccount.createAccount()
+    await sleep(2000)
+    await marginAccount.refresh()
+
     const instructions: TransactionInstruction[] = []
-    await marginAccount.withCreateAccount(instructions)
-    const position = await marginAccount.withGetOrCreatePosition({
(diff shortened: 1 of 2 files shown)


libraries/ts/src/margin/pool/pool.ts
```diff
@@ -446,8 +446,9 @@ export class Pool {
     await marginAccount.createAccount()
     await sleep(2000)
     await marginAccount.refresh()
+
     const instructions: TransactionInstruction[] = []
-    const position = await marginAccount.withGetOrCreatePosition(this.addresses.depositNoteMint)
+    const position = await marginAccount.getOrCreatePosition(this.addresses.depositNoteMint)
```
Tristyn, line 451: This is going back to `getOrCreatePosition`. Can this be `withGetOrCreatePosition`?

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -648,11 +652,13 @@ export class Pool {
...
-    const depositPosition = await marginAccount.withGetOrCreatePosition(this.addresses.depositNoteMint)
-    assert(depositPosition)
-
     const refreshInstructions: TransactionInstruction[] = []
     const instructions: TransactionInstruction[] = []
+    const depositPosition = await marginAccount.withGetOrCreatePosition({
+      positionTokenMint: this.addresses.depositNoteMint,
+      instructions: refreshInstructions
+    })
+    assert(depositPosition)
```
Tristyn, line 661: Perfect

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -792,7 +798,7 @@ export class Pool {
     destination?: TokenAddress
   }) {
     // FIXME: can source be calculated in withdraw?
-    const source = await marginAccount.withGetOrCreatePosition(this.addresses.depositNoteMint)
+    const source = await marginAccount.getOrCreatePosition(this.addresses.depositNoteMint)
```
Tristyn, line 801: This is also going back to getOrCreatePosition. Can this be withGetOrCreatePosition?

Tristyn (commented):
There's 2 instances where we stop using the `with` function. Is that done to get tests to pass? And if so, those tests need to get fixed

Edit: After checking DMs, yes using `with` makes the test fail. hmmm

Tristyn (approved):
This PR gives functions more correct names, and using `with` functions can happen in a later PR

PR #165, opened 2022-07-15
place and cancel serum orders, increase feeCover from 0.02 to 0.07 SOL

loganjross:


libraries/ts/src/margin/marginAccount.ts
```diff
@@ -881,6 +894,203 @@ export class MarginAccount {
...
+    selfTradeBehavior = "decrementTake",
+    clientOrderId = new BN(Date.now()),
+    payer = this.address
+  }: {
+    market: MarginMarketConfig
+    orderSide: "sell" | "buy" | "ask" | "bid"
+    orderType: "limit" | "ioc" | "postOnly"
+    orderPrice: number
+    orderSize: TokenAmount
+    selfTradeBehavior: "decrementTake" | "cancelProvide" | "abortTransaction"
```
Tristyn, line 939: Can you create types for `"decrementTake" | "cancelProvide" | "abortTransaction"` and the others

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -881,6 +894,203 @@ export class MarginAccount {
...
+
+    // If trading on margin
+    if (orderAmount.gt(accountPoolPosition.depositBalance) && this.pools) {
+      const difference = orderAmount.sub(accountPoolPosition.depositBalance)
+      const pool = this.pools[market.baseSymbol as MarginPools]
+      await pool.marginBorrow({
+        marginAccount: this,
+        pools: Object.values(this.pools),
+        change: PoolTokenChange.setTo(accountPoolPosition.loanBalance.add(difference))
+      })
```
Tristyn, line 955: Only problem with this being included is it will cause the wallet to ask for approval more than once

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -881,6 +894,203 @@ export class MarginAccount {
...
+    payer: PublicKey
+  }) {
+    const instructions: TransactionInstruction[] = []
+    const orderAmount = orderSize.divn(orderPrice)
+    const accountPoolPosition = this.poolPositions[market.baseSymbol as MarginPools]
+
+    // If trading on margin
+    if (orderAmount.gt(accountPoolPosition.depositBalance) && this.pools) {
+      const difference = orderAmount.sub(accountPoolPosition.depositBalance)
+      const pool = this.pools[market.baseSymbol as MarginPools]
```
Tristyn, line 950: What if pool is undefined?

Tristyn (dismissed):
Can you move this serum stuff to the src/margin/serum folder. As we add adapters the margin account will get very cluttered. Because the methods are getting moved out of the margin account they'll need the margin account as a parameter

src/margin/serum/market.ts might be a good candidate if you don't have too much trouble instantiating it. market.ts needs a facelift though, there's extra crud in there that could be done simpler. All the program ID fields in it could be pulled from config instead for example

libraries/ts/src/margin/config.ts
```diff
@@ -4,7 +4,7 @@ import MARGIN_CONFIG from "./config.json"
 export type MarginTokens = "BTC" | "ETH" | "MSRM" | "SOL" | "SRM" | "USDC"
 export type MarginOracles = "BTC_USD" | "ETH_USD" | "SOL_USD" | "SRM_USD"
 export type MarginPools = "BTC" | "ETH" | "SOL" | "SRM" | "USDC"
-export type MarginMarkets = "BTC_USDC" | "ETH_USDC"
+export type MarginMarkets = "BTC/USDC" | "ETH/USDC"
```
Tristyn, line 7: These can't be renamed as they're keys in MarginConfig
```markets: Record<MarginMarkets, MarginMarketConfig>```
Must deploy a config.json that uses these configs, then these can be renamed

File: libraries/ts/src/margin/config.ts [hunk 5]
```diff
@@ -65,8 +65,8 @@ export interface MarginMarketConfig {
   quoteSymbol: MarginTokens
   requestQueue: Address
   eventQueue: Address
-  bids: Address
-  asks: Address
+  bidsAddress: Address
+  asksAddress: Address
```
Tristyn, line 69: Can't rename these either unless it's renamed in the initializer as well

libraries/ts/src/margin/marginClient.ts
```diff
@@ -74,7 +74,7 @@ export class MarginClient {
 
   static getConfig(cluster: MarginCluster): MarginConfig {
     if (typeof cluster === "string") {
-      return JET_CONFIG[cluster] as MarginConfig
+      return JET_CONFIG[cluster] as unknown as MarginConfig
```
Tristyn, line 77: It needs to compile sir. Your interfaces don't match the json the initializer created. You can get it to compile here, but if you use it in a frontend it'll throw errors there anyway. Don't turn off the typechecker
```suggestion
      return JET_CONFIG[cluster] as MarginConfig
```

libraries/ts/src/margin/serum/market.ts
```diff
@@ -1,785 +1,421 @@
...
-      ts: number
-    }
-  }
-
-  private marginSerumProgramId: PublicKey
-  private marginSerumAdapterMetadata: PublicKey
+  provider: AnchorProvider
+  programs: MarginPrograms
+  marketConfig: MarginMarketConfig
+  serum: SerumMarket
```
Tristyn, line 33: All these fields are good. I don't think the market deserves a provider because it's not tied to a specific user. MarginAccount get's a provider because the provider wallet literally owns the margin account and can sign to change it
```suggestion
  programs: MarginPrograms
  marketConfig: MarginMarketConfig
  serum: SerumMarket
```

libraries/ts/src/margin/serum/market.ts
```diff
@@ -1,785 +1,421 @@
...
-      (await PublicKey.findProgramAddress([marginSerumProgramId.toBuffer()], metadataProgramId))[0],
-      layoutOverride
-    )
+  get name(): string {
+    return this.marketConfig.symbol
   }
-
   get address(): PublicKey {
-    return this._decoded.ownAddress
+    return new PublicKey(this.marketConfig.market)
```
Tristyn, line 39: `translateAddress` is a bit better than `new PublicKey` here

Tristyn (changes requested):
Please undo changes to config.ts, remove the `as unknown` cast, and resolve conflicts so the github checks can run

One thing I'm worried about is this might break Mas' market making bot, if that repo were added as a git submodule we'd be able to include it in tests

Tristyn (approved):

PR #169, opened 2022-07-18
add liquidationEndingCollateral to PoolPosition

loganjross:


libraries/ts/src/margin/marginAccount.ts
```diff
@@ -317,6 +318,12 @@ export class MarginAccount {
       // Max trade amounts
       const maxTradeAmounts = this.getMaxTradeAmounts(pool, depositBalance, loanBalance)
 
+      // Minimum amount to deposit for the pool to end a liquidation
+      const liquidationEndingCollateral = this.valuation.requiredCollateral
+        .asTokenAmount(pool.decimals)
+        .sub(this.valuation.effectiveCollateral.asTokenAmount(pool.decimals).muln(MarginAccount.RISK_WARNING_LEVEL))
+        .div(pool.depositNoteMetadata.valueModifier.asTokenAmount(pool.decimals).muln(MarginAccount.RISK_WARNING_LEVEL))
+
```
Tristyn, line 326: Can you call `asTokenAmount` once at the end?

There is also a divide by zero error, can you check for diving by zero?

Tristyn (changes requested):
Can you call asTokenAmount once at the end?

There is also a divide by zero error, can you check for diving by zero so tests pass?

Tristyn (approved):

PR #171, opened 2022-07-18
Idl gen

Tristyn:
Just playing around with the idea of having json IDL files that are merged with what's in /target/idl. So our extra type definitions are included there.

The ts lib could automatically receive updated IDLs as part of anchor build, and return an error if a type definition is required but missing


Tristyn:
I might revisit this in the future

PR #182, opened 2022-07-20
stop tracking token account balances with MarginPool

dnut:
use the token accounts/mints as needed. aggregate dependencies and execute calculations with a declarative builder api that checks dependencies at compile time

a few instructions require more accounts to get the sufficient data, but in most cases it was all there already.

### DIFF
--- a/programs/margin-pool/src/events.rs
+++ b/programs/margin-pool/src/events.rs
@@ -92,9 +92,6 @@ pub struct Collect {
 pub struct MarginPoolSummary {
     pub borrowed_tokens: u64,
     pub uncollected_fees: u64,
-    pub deposit_tokens: u64,
-    pub deposit_notes: u64,
-    pub loan_notes: u64,
     pub accrued_until: i64,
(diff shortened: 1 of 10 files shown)


dnut:
I'm seeing a bug in our current master where the rounding is done too many times. It's being overly defensive with consecutive roundings, and I think the changes here would address it
```
    Program log: TokenChange { kind: ShiftBy, tokens: 1 }
    Program log: FullAmount { tokens: 1, notes: 2 }
    Program log: FullAmount { tokens: 1, notes: 0 }
```
It's not a security vulnerability. It's the opposite. We're losing too much precision out of an over-abundance of paranoia, which means a borrow of 1 token is impossible.

I could split out my amount conversion refactoring into another PR but that'll be some work. it would also be some work to update this PR with the latest code, but probably less.

dnut:
this is very stale and probably never going to be integrated in full. if we want any bits and pieces of this, they can be manually copied

PR #185, opened 2022-07-20
JV15-223 Add checks for various SOL contingencies

rhubarbgarden:
1. If the SOL wallet balance is > 0.07 greater than the borrowed amount, `maxInput.repay` should be the same as the `loanBalance`

2. If the wallet balance is <= 0.07 SOL greater than, or less than, the borrowed amount, `maxInput.repay` should be `wallet balance - feeCover`

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,7 @@
-* @suscd @partiallysorted @nevi-me @dnut @callensm @jetjake @tristyn
-.dockerignore @callensm @suscd @jetjake @davodesign84
-.github/ @callensm @suscd @jetjake @davodesign84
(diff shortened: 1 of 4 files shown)


Tristyn:
I pushed a commit that simplifies the fee cover. It subtracts 0.07 from the sol wallet balance at the start. It was too big to fit in a github suggestion. If that's not ideal you can just revert the commit

Tristyn (approved):

Tristyn (changes requested):
I don't think it needs this extra commit [3a1af9f](https://github.com/jet-lab/jet-v2/pull/185/commits/3a1af9f980d8aa49bc6ebbcb43720ea00066fba9) on top

Tristyn (approved):

PR #193, opened 2022-07-22
Removes fee collection threshold

eyevz:
We have been using a fee collection threshold to ensure that rounding does not cost us an appreciable fraction of fees collected. We can instead forgo the threshold and explicitly track the remainder as we collect fees.

This removes one parameter per pool for us to manage.

### DIFF
--- a/programs/margin-pool/src/state.rs
+++ b/programs/margin-pool/src/state.rs
@@ -573,7 +573,8 @@ pub struct MarginPoolConfig {
     /// The fee rate applied to interest payments collected
     pub management_fee_rate: u16,
 
-    _reserved: u64
+    /// Unused
+    pub reserved: u64
 }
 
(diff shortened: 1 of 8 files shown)


programs/margin-pool/src/state.rs
```diff
@@ -570,9 +572,6 @@ pub struct MarginPoolConfig {
 
     /// The fee rate applied to interest payments collected
     pub management_fee_rate: u16,
-
-    /// The threshold for fee collection
-    pub management_fee_collect_threshold: u64,
```
suscd, line 575: best not to change the size of the account, keep the now unused field and change the name to `_reserved` instead.

suscd (changes requested):

Tristyn (commented):
If there's no state or instruction api changes then often no lib changes are required

PR #199, opened 2022-07-23
less-tx-approvals

loganjross:


Tristyn (dismissed):
Great job. Tests just need to pass

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -55,7 +54,7 @@ export interface PoolProjection {
   borrowRate: number
 }
 
-const feesBuffer: number = LAMPORTS_PER_SOL * 0.02
+const feesBuffer: number = LAMPORTS_PER_SOL * 0.075
```
Tristyn, line 57: Maybe feesBuffer can be combined with feeCover from `maxTradeAmounts`?

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -803,13 +802,18 @@ export class Pool {
...
     const preInstructions: TransactionInstruction[] = []
     const refreshInstructions: TransactionInstruction[] = []
     const instructions: TransactionInstruction[] = []
 
+    let source = marginAccount.getPosition(this.addresses.depositNoteMint)?.address
+    if (!source) {
+      source = await marginAccount.withGetOrCreatePosition({
+        positionTokenMint: this.addresses.depositNoteMint,
+        instructions: preInstructions
+      })
```
Tristyn, line 814:
```suggestion
      throw new Error("No deposit position")
```

Tristyn (approved):
Looks good just a few suggestions

libraries/ts/src/margin/pool/pool.ts
Tristyn: If `change` is trying to set the deposit position then this could return early because it's already less than what you're setting it to and we want that to be a no-op in the program if the transaction were to be sent. If `change` is trying to shift it I think it should throw. But that's not completely necessary

libraries/ts/src/margin/marginAccount.ts
```diff
@@ -375,16 +375,15 @@ export class MarginAccount {
...
 
+    // Wallet's balance for pool
+    // If depsiting or repaying SOL, maximum input should consider fees
     let walletAmount = (pool.symbol && this.walletTokens?.map[pool.symbol].amount) ?? TokenAmount.zero(pool.decimals)
-
-    // If depsiting or repaying SOL, maximum input should still cover fees
     if (pool.tokenMint.equals(NATIVE_MINT)) {
-      const feeCover = TokenAmount.tokens(0.07, pool.decimals)
-      walletAmount = TokenAmount.max(walletAmount.sub(feeCover), TokenAmount.zero(pool.decimals))
+      walletAmount = TokenAmount.max(walletAmount.subn(feesBuffer), TokenAmount.zero(pool.decimals))
```
Tristyn, line 382: `subn` `muln` and all the other number functions are very unreliable when numbers get big
```suggestion
      walletAmount = TokenAmount.max(walletAmount.sub(numberToBN(feesBuffer)), TokenAmount.zero(pool.decimals))
```

Tristyn (dismissed):
Can you replace the `subn` call with `sub` and we're good to go

Tristyn (approved):

PR #210, opened 2022-07-27
Margin actions return txid as string

rhubarbgarden:


libraries/ts/src/utils/sendAll.ts
```diff
@@ -50,21 +50,26 @@ export async function sendAll(
...
+      const txnArray = await Promise.all(
         transactions.map(async tx => {
           if (tx.instructions.length > 0) {
             const rawTx = tx.serialize()
             return await sendAndConfirmRawTransaction(provider.connection, rawTx, opts)
           }
         })
       )
+      // Return the txid of the final transaction in the array
+      lastTxn = txnArray[txnArray.length - 1] ?? ""
```
jetjake, line 67: Why would we send the last txn and not a list of them?

jetjake (commented):
One inline question.

Also, how have you tested this?

libraries/ts/src/utils/sendAll.ts
rhubarbgarden: We could do it either way. The relevant txid that we would want to get for the purposes of identifying the correct block explorer URL would be the final in the array, so we could either:
- Return the final txid from jet-v2, or
- Return the entire array from jet-v2 and find the final one from the UI.

The first way seems cleaner to me since when we should only be opening one new page when clicking on the notification component, corresponding to one tx.

rhubarbgarden:
This was tested by running `npm run test` within this repo as well as running `npm run debug` from the jet-app repo and manually testing the expected function of the notification component.

libraries/ts/src/utils/sendAll.ts
jetjake: Here's my problem with that, it seems like we're trying to solve a problem specific to block explorer (a frontend feature) in the library.  The library is not simply a convenience for the frontend, it is an API that is used for many different use cases and should have its own standard/design. With that said, we need to do some work on the client library right after this launch, so we can merge this for now.  Please add a "TODO in the code that says we should return an array and then we can merge this.

davodesign84 (approved):

jetjake (dismissed):

PR #215, opened 2022-07-29
Miscellaneous changes (update events)

nevi-me:
* Makes margin::events public
* Remove unused variables

### DIFF
--- a/programs/margin-pool/src/events.rs
+++ b/programs/margin-pool/src/events.rs
@@ -12,6 +12,8 @@ pub struct PoolCreated {
     pub authority: Pubkey,
     pub payer: Pubkey,
     pub summary: MarginPoolSummary,
+    pub version: u8,
+    pub pool_bump: u8,
 }
 
 #[event]
--- a/programs/margin-pool/src/events.rs
+++ b/programs/margin-pool/src/events.rs
@@ -13,7 +13,6 @@ pub struct PoolCreated {
     pub payer: Pubkey,
     pub summary: MarginPoolSummary,
     pub version: u8,
-    pub pool_bump: u8,
 }
 
 #[event]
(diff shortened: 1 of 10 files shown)


eyevz:
Looks sensible to me.

programs/margin-pool/src/events.rs
```diff
@@ -12,6 +12,8 @@ pub struct PoolCreated {
     pub authority: Pubkey,
     pub payer: Pubkey,
     pub summary: MarginPoolSummary,
+    pub version: u8,
+    pub pool_bump: u8,
```
suscd, line 16: why include this?

programs/margin/src/events.rs
```diff
@@ -56,6 +56,7 @@ pub struct PositionBalanceUpdated {
 
 #[event]
 pub struct PositionTouched {
+    pub margin_account: Pubkey,
```
suscd, line 59: why is this necessary? The margin account is already recorded in the invoke event, including it here also for every single position seems annoyingly redundant (given log size is already an issue).

eyevz: No objection. There are certainly savings to be had if we're willing to grab `Pubkey` fields from the instruction calls.

programs/margin/src/instructions/register_position.rs
```diff
@@ -82,7 +82,6 @@ pub fn register_position_handler(ctx: Context<RegisterPosition>) -> Result<()> {
     let position = account.get_position_by_key(&key).require()?;
 
     emit!(events::PositionRegistered {
-        margin_account: ctx.accounts.margin_account.key(),
```
suscd, line 85: A potential issue with removing this is you don't know the margin account for the case where users directly register a position (as opposed to through invoke, which records an extra event that mentions the margin account).

suscd (approved):

PR #222, opened 2022-08-01
Remove set-to-nonzero

eyevz:
Addresses [JV15-276](https://jetprotocol.atlassian.net/browse/JV15-276), and should fixed the dust in the number of tokens delivered upon borrow, when a deposit of the same asset already exists, @davodesign84

### DIFF
--- a/libraries/ts/src/margin/pool/pool.ts
+++ b/libraries/ts/src/margin/pool/pool.ts
@@ -521,7 +521,9 @@ export class Pool {
     change: PoolTokenChange
     destination?: TokenAddress
   }): Promise<string> {
-    assert(change.changeKind.isShiftBy())
+    if (!change.changeKind.isShiftBy()) {
+      throw new Error('Use ShiftBy for all borrow instructions')
+    }
 
     await marginAccount.refresh()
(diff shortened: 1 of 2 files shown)


libraries/ts/src/margin/pool/pool.ts
```diff
@@ -521,6 +521,8 @@ export class Pool {
     change: PoolTokenChange
     destination?: TokenAddress
   }): Promise<string> {
+    assert(change.changeKind.isShiftBy())
```
davodesign84, line 524: `assert` is a testing library, and doesn't play well with typescript for type inference. This should be changed to something like 
```
if (!change.changeKind.isShiftBy()) {
throw new Error('Error description')
}
```

davodesign84 (changes requested):
Looks good except for the `assert` usage :)

libraries/ts/src/margin/pool/pool.ts
eyevz: Done.

davodesign84 (approved):

jetjake (approved):

jetjake (approved):

PR #224, opened 2022-08-01
Restrict SetTo behaviour to set-to-zero

eyevz:
SetTo with non-zero token values is problematic. We have modified the front end to use only set-to-zero behaviour and now want to enforce this in the margin pool program.

JV15-276

We would ideally refactor the program api further.

### DIFF
--- a/programs/margin-pool/src/lib.rs
+++ b/programs/margin-pool/src/lib.rs
@@ -109,7 +109,7 @@ pub use detail::TokenChange;
 mod detail {
     use anchor_lang::prelude::*;
 
-    use crate::{ChangeKind, Amount, ErrorCode};
+    use crate::{Amount, ChangeKind, ErrorCode};
 
     /// Interface for changing the token value of an account through pool instructions
     #[derive(AnchorSerialize, AnchorDeserialize, Debug, Clone, Copy)]
--- a/programs/margin-pool/src/lib.rs
+++ b/programs/margin-pool/src/lib.rs
@@ -105,41 +105,48 @@ mod jet_margin_pool {
     }
 }
 
-/// Interface for changing the token value of an account through pool instructions
-#[derive(AnchorSerialize, AnchorDeserialize, Debug, Clone, Copy)]
-pub struct TokenChange {
-    kind: ChangeKind,
-    tokens: u64,
-}
+pub use detail::TokenChange;
+mod detail {
+    use anchor_lang::prelude::*;
 
-impl TokenChange {
-    pub fn new(kind: ChangeKind, tokens: u64) -> Self {
-        Self { kind, tokens }
+    use crate::{ChangeKind, Amount};
+
+    /// Interface for changing the token value of an account through pool instructions
+    #[derive(AnchorSerialize, AnchorDeserialize, Debug, Clone, Copy)]
+    pub struct TokenChange {
+        kind: ChangeKind,
+        tokens: u64,
     }
 
-    pub const fn set(value: u64) -> Self {
-        Self {
-            kind: ChangeKind::SetTo,
-            tokens: value,
+    impl TokenChange {
+        pub fn new(kind: ChangeKind, tokens: u64) -> Self {
+            Self { kind, tokens }
         }
-    }
-    pub const fn shift(value: u64) -> Self {
-        Self {
-            kind: ChangeKind::ShiftBy,
-            tokens: value,
+
+        pub const fn set(value: u64) -> Self {
+            Self {
+                kind: ChangeKind::SetTo,
+                tokens: value,
+            }
         }
-    }
+        pub const fn shift(value: u64) -> Self {
+            Self {
+                kind: ChangeKind::ShiftBy,
+                tokens: value,
+            }
+        }
 
-    pub fn amount(&self) -> Amount {
-        Amount::tokens(self.tokens)
-    }
+        pub fn amount(&self) -> Amount {
+            Amount::tokens(self.tokens)
+        }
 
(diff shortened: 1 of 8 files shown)


programs/margin-pool/src/lib.rs
```diff
@@ -108,11 +108,15 @@ mod jet_margin_pool {
 /// Interface for changing the token value of an account through pool instructions
 #[derive(AnchorSerialize, AnchorDeserialize, Debug, Clone, Copy)]
 pub struct TokenChange {
-    pub kind: ChangeKind,
-    pub tokens: u64,
+    kind: ChangeKind,
+    tokens: u64,
```
eyevz, line 112: Even though I've made these fields private, TokenChange can still be initialised directly, [like this](https://github.com/jet-lab/jet-v2/blob/607800749b938b642a394ec5704f55a28cab13a8/programs/margin-pool/src/instructions/withdraw.rs#L89).

eyevz: This is because `TokenChange` is at crate level so the instruction is essentially a submodule. In the next commit, by branching `TokenChange` into a submodule, we remove access to private members from the instructions.

qiqicodes (approved):
lgtm =]

dnut (commented):
I don't see why this is needed in the program. It's a UX decision for the frontend. This functionality could still be useful for cranks. Unless there is a security issue with allowing someone to set to a non-zero amount, I don't think this belongs here.

Also, if 0 is the only valid value for SetTo, I would have probably implemented this using the type system rather than by adding checks.
```
enum TokenChange {
    SetToZero,
    ShiftBy(u64)
}
```

suscd (changes requested):
I agree with Drew, this change seems unnecessary for the program. What compelling reason is there to do this? The API is not fundamentally broken, the way I see it, it just has some behaviour you might see as undesirable, but that seems more like a matter of perspective than a core problem that needs to be avoided. 

So if it's a problem for the frontend, the frontend can simply just elect to not use this. 

To be clear I'm not against improving the API, but this does not look like an improvement.

eyevz:
> I don't see why this is needed in the program. It's a UX decision for the frontend. This functionality could still be useful for cranks. Unless there is a security issue with allowing someone to set to a non-zero amount, I don't think this belongs here.
> 
> Also, if 0 is the only valid value for SetTo, I would probably implement this with types instead of runtime checks.
> 
> ```
> enum TokenChange {
>     SetToZero,
>     ShiftBy(u64)
> }
> ```

That's a fair point. My I wanted to make the smallest possible change.

I can't see a case where you'd ever want to SetTo non-zero token amount. The whole purpose was to introduce idempotent instructions, which these because the token value of deposits is not constant.

dnut:
> I do think there is something fundamentally wrong with the api though.

I don't see anything inherently problematic with the concept of setting your account to a specific balance. If it results in unexpected balances, that's probably because of a counter-intuitive approach to rounding.

> I can't see a case where you'd ever want to SetTo a non-zero token amount. Imo it will only ever be a foot-gun.

- The user submits a transaction but doesn't see the transaction confirmed soon enough and they assume it failed. So they execute the same action again moments later. But actually both transactions end up executing successfully.
- A trading bot wants to be sure that a transaction executes as soon as possible during poor network conditions. It's not acceptable to wait for every transaction to timeout and try again if it failed. The bot administrator is happy to pay extra transaction fees to send 10 transactions at a time to ensure at least one is confirmed.

> The whole purpose was to introduce idempotent instructions, which these are not, because the token value of deposits is not constant.

The operation is idempotent because if you apply it to its own output, the next output is identical. If you apply the operation to a different starting value, it's true that you'll see a distinct behavior. I agree there are other effects causing gradual state changes constantly, so it won't apply directly to its own output unless it is duplicated within the same slot. I don't really see the problem here though. The differences from one slot to the next are negligible. The idea was that if you send duplicates in a short period of time it will have basically the same effect regardless of the number of executions. Are these tiny changes between slots causing a problem? Or is it the rounding of conversions between notes and tokens?

dnut:
> @dnut This is well beyond defensive rounding:

I agree. There are more decisions about precision than just being defensive. But the fundamental problem is with precision. There are arbitrary decisions about what numbers will be affected by how many stages of rounding and in which direction, independent of defensiveness.

> By eliminating set-to-nonzero these now exhibit single lamport discrepancies only.

With ShiftBy, you bypass the bug. But we can fix the bug. And even improve the behavior for ShiftBy.

What we need to do is look at each instruction and decide what matters. Where can precision be compromised, and what is critical to meet user expectations. Then make sure the calculation meets those requirements. We probably need to give specified token amounts more priority than we give them now.

For example, when a user wants to set their borrow to 10 USDC, or they want to shift their borrow by an additional 10 USDC, what does that mean? We need to enumerate the exact priorities so we can meet them with any implementation (either set-to or shift-by). Here's an example (my guess) at the priorities:
- It's critical to give them at least 10 USDC.
- It's OK to give them slightly more.
- It's OK if they owe slightly more than 10 USDC after borrowing.

There are two kinds of borrows that could be implemented for this:
1. Direct borrow to wallet: Transfer 10 USDC direct to their wallet. Calculate the number of loan notes this corresponds to, rounding up, and transfer that to their claim position.
2. Borrow deposit notes: Calculate the number of deposit notes that equals 10 USDC. Round *up* to make sure they get enough, and transfer that many deposit notes to their margin account. Then they will definitely be able to withdraw at least 10 USDC. Calculate the amount of USDC the deposit notes correspond to, and convert it to loan notes, then round up to determine the size of the claim.

PR #230, opened 2022-08-03
Improved risk indicator

eyevz:
This risk indicator is a bit nicer than the one we've been using. It is defined as

`indicator = (liabilities + required_collateral) / weighted_collateral`

which is a leverage-like quantity but in the weighted space. The threshold for liquidation is `indicator > 1.0`.

This indicator has the nice property that `1.0 - indicator` is the return on collateral that will push an account to the liquidation threshold.

E.g. If I have SOL collateral and a USDC loan, and my risk indicator is 0.88, that means a 12% decline in the SOL price will bring  my account to the liquidation threshold.

----

This PR also enables the setup limit on the UI, preventing users from setting up to the liquidation threshold.


Tristyn (approved):

Tristyn:
> E.g. If I have SOL collateral and a USDC loan, and my risk indicator is 0.82, that means a 12% decline in the SOL price will bring my account to the liquidation threshold.

This property is really nice

PR #231, opened 2022-08-04
add `get_pools` utility for finding swap pools in rust sdk

qiqicodes:
port the `get_pools` from liquidator crank to rust sdk. this will find all available swap pools for supported mints.

- [x] ports `get_pools` to rust sdk
   `get_pools` parses swap pools and return pools with most liquidity
- [x] updates cargo.toml
- [x] adds helper functions
- [x] adds local and mainnet tests


libraries/rust/src/swap.rs
```diff
@@ -18,11 +18,21 @@
 //! The margin swap module allows creating simulated swap pools
 //! to aid in testing margin swaps.
```
suscd, line 19: This seems inaccurate, can you also update this?

This entire module should just be renamed to `spl_swap`, because there are other swap pools that this does not support.

libraries/rust/src/swap.rs
```diff
@@ -18,11 +18,21 @@
...
+use anchor_lang::AccountDeserialize;
+use anchor_spl::token::spl_token::state::Mint;
+use anyhow::Result;
+use jet_proto_math::Number128;
 use jet_simulation::solana_rpc_api::SolanaRpcClient;
+use parking_lot::RwLock;
+use pyth_sdk_solana::PriceFeed;
 use solana_sdk::{program_pack::Pack, pubkey::Pubkey};
 use spl_token_swap::state::SwapV1;
+use tracing::*;
```
suscd, line 33: don't glob import tracing

libraries/rust/src/swap.rs
```diff
@@ -41,42 +51,151 @@ impl SwapPool {
     /// Get all swap pools that contain pairs of supported mints
     pub async fn get_pools(
         rpc: &Arc<dyn SolanaRpcClient>,
-        supported_mints: HashSet<Pubkey>,
+        supported_mints: &HashMap<Pubkey, Mint>,
         swap_program: Pubkey,
-    ) -> anyhow::Result<Vec<Self>> {
+        price_cache: PriceCache,
```
suscd, line 56: For this to be useful for anyone, we have to also provide the methods to construct the `supported_mints` and `price_cache`. How is a user supposed to get those?

suscd (dismissed):

libraries/rust/src/swap.rs
qiqicodes: removed `price_cache` and revert `supported_mints` data structures for `get_pools`

libraries/rust/src/spl_swap.rs
```diff
@@ -0,0 +1,227 @@
...
+    use solana_sdk::signature::Keypair;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_mainnet_orca_pools() -> anyhow::Result<()> {
+        let payer = Keypair::new();
+        let rpc: Arc<dyn SolanaRpcClient> = Arc::new(RpcConnection::new_optimistic(
+            payer,
+            "https://api.mainnet-beta.solana.com",
```
suscd, line 200: you shouldn't add a unit test that has external network dependencies, like mainnet. You can just remove this test, and rely on the simulator tests to exercise this (which it looks like it does).

suscd (approved):

PR #236, opened 2022-08-09
fix undefined wallet map bug

loganjross:
When getting maxTradeAmounts, if this.walletTokens is undefined then we are thrown an error at: this.walletTokens.map[pool.symbol].amount

### DIFF
--- a/libraries/ts/src/margin/marginAccount.ts
+++ b/libraries/ts/src/margin/marginAccount.ts
@@ -395,7 +395,7 @@ export class MarginAccount {
     // If depsiting or repaying SOL, maximum input should consider fees
     let walletAmount = TokenAmount.zero(pool.decimals)
     if (pool.symbol && this.walletTokens) {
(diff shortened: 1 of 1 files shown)


libraries/ts/src/margin/marginAccount.ts
```diff
@@ -401,7 +394,8 @@ export class MarginAccount {
     // Wallet's balance for pool
     // If depsiting or repaying SOL, maximum input should consider fees
     let walletAmount =
-      (pool.symbol ? this.walletTokens?.map[pool.symbol].amount : undefined) ?? TokenAmount.zero(pool.decimals)
+      (pool.symbol && this.walletTokens ? this.walletTokens.map[pool.symbol].amount : undefined) ??
+      TokenAmount.zero(pool.decimals)
```
Tristyn, line 398: Can this just be an if statement instead?
I have to look at it for a while to understand what it's doing

Tristyn (approved):

PR #237, opened 2022-08-09
github workflow refactor: faster + reproducible

dnut:
example execution: https://github.com/jet-lab/jet-v2/actions/runs/2842020855

Github workflow now runs in a docker image containing all rust+node+solana+anchor dependencies: https://github.com/jet-lab/dockerfiles/blob/master/Dockerfile

You can run the workflow locally:
```bash
./check all
```
and in docker
```bash
./check in-docker all
```
or run a single job, with or without docker
```bash
./check in-docker anchor-test
```

## code reuse + abstraction
- rather than having to install dependencies within github workflow, the workflow now executes in a standardized reusable docker image that we can share across many projects.
- github workflow can focus on its strengths: concurrency, caching, and integration with git and other ci tools, rather than acting as a convoluted shell script
- inevitable shell steps encapsulated within `check` shell script to run verifications

## faster builds
Once cached, it completes in about 10.5 minutes, which is about an 8 minute improvement over the prior execution time. still not ideal, but it's hard to make these super fast because there's a lot that actually needs to run for every commit.

## less coupled with github
Previously, everything was thrown into the github workflow, which meant that any time you wanted to reproduce any CI behavior outside of github, you needed to carefully read the complicated series of steps in the workflow yaml, and manually translate them to whatever steps and syntax are needed for the unique non-standard environment you happened to be using for development.

Now:
- the environment is standardized - the docker image is pre-packaged with solana dependencies and can run anywhere.
- The workflow simply executes `check`, which means:
  - CI failures are easily reproduced locally with `./check <job-name>` or `./check in-docker <job-name>`
  - workflow can be easily migrated to other CI tools

### naming
I renamed two of the jobs, and I split an existing job into two:

| old | new |
| --- | --- |
| localnet-tests | hosted-tests-localnet |
| 'Typescript Integration Tests' | anchor-test |
| coverage | cargo-test<br>cargo-lint |

The motivation was twofold:
- Standardize names between the github workflow and the `check` script. They need to be bash function names, so "Typescript Integration Tests" won't do.
- Expressiveness. I believe the new names offer greater transparency into what the job actually does.

So the merge-to-master requirements in the repo config need to be changed, which I do not have permission to modify. The three old names need to be removed, and the four new names need to be added.

### could be faster
The localnet tests are the bottleneck. The script takes about 7.5 minutes to run even when all dependencies are cached. The remaining 3 minutes are for setup/teardown of the github job (similar for each job) which I don't see a way around.

There is one way to optimize these tests. Break them into chunks and execute them in parallel. If we use two chunks, it would cut the build time down by 3.5 minutes, so we'd be looking at ~7 minute execution time instead of 10.5.

The trade-off here is complexity. The workflow would have at least two localnet tests: localnet batch 1 and localnet batch 2. Then the check and test_on_localnet scripts would need additional complexity to execute either half or all tests based on configuration.

I couldn't decide if the trade-off is worthwhile, curious on opinions here. This could always go in a future PR after this is merged.


jetjake (dismissed):
Hell yeah.

suscd (changes requested):
The anchor test doesn't look like it works:

https://github.com/jet-lab/jet-v2/runs/7793520103?check_suite_focus=true#step:5:107

dnut:
@suscd should be fixed now

suscd (approved):

PR #242, opened 2022-08-13
integrate-pool-swaps

loganjross:


libraries/ts/src/margin/pool/pool.ts
```diff
@@ -873,6 +881,204 @@ export class Pool {
...
+    } else {
+      // Pedantic. We can't reach this condition if correct pool is selected
+      throw new Error("Invalid swap pool selected")
+    }
+
+    // Swap
+    await marginAccount.withAdapterInvoke({
+      instructions,
+      adapterProgram: this.programs.config.marginSwapProgramId,
+      // TODO: check if this evaluates to DUheebnZ below
```
davodesign84, line 1022: Is this intended to be part of this PR?

davodesign84 (dismissed):
Can we please add some Anchor tests when we add new functionality? :)

libraries/ts/src/margin/pool/pool.ts
loganjross: yes

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -873,6 +881,212 @@ export class Pool {
...
+    const destinationAccount = await marginAccount.withGetOrCreatePosition({
+      positionTokenMint: outputToken.addresses.depositNoteMint,
+      instructions
+    })
+
+    // Loan note account
+    const loanNoteAccount = await marginAccount.withGetOrCreatePosition({
+      positionTokenMint: this.addresses.loanNoteMint,
+      instructions
+    })
```
Tristyn, line 947: This is only required when borrowing

It can be moved inside the `if` statement below. I checked usage and `loanNoteAccount` is only used in the `withMarginBorrow` call so they can be put together

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -873,6 +881,212 @@ export class Pool {
...
+
+    // If swapping on margin
+    const accountPoolPosition = marginAccount.poolPositions[this.symbol ?? ""]
```
Tristyn, line 950: assert `this.symbol` at the start so you don't have to do this `this.symbol ?? ""` weirdness

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -873,6 +881,212 @@ export class Pool {
...
+      const transitSourceAccountIx = createAssociatedTokenAccountInstruction(
+        marginAccount.addresses.owner,
+        transitSourceAccount,
+        marginAccount.address,
+        this.addresses.tokenMint,
+        TOKEN_PROGRAM_ID,
+        ASSOCIATED_TOKEN_PROGRAM_ID
+      )
+      instructions.push(transitSourceAccountIx)
+    }
```
Tristyn, line 976: `AssociatedToken.withCreate` can replace everything in this block of code

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -873,6 +881,212 @@ export class Pool {
...
+      const transitDestinationAccountIx = createAssociatedTokenAccountInstruction(
+        marginAccount.addresses.owner,
+        transitDestinationAccount,
+        marginAccount.address,
+        outputToken.addresses.tokenMint,
+        TOKEN_PROGRAM_ID,
+        ASSOCIATED_TOKEN_PROGRAM_ID
+      )
+      instructions.push(transitDestinationAccountIx)
+    }
```
Tristyn, line 995: `AssociatedToken.withCreate` same here

Tristyn, line 995: `AssociatedToken.withCreate` same here

Tristyn (approved):
The code can be simplified
* The borrow position can be registered only if borrowing is required
* Checking and creating an ATA can be done with one call to `AssociatedToken.withCreate`

But functional none the less

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -940,16 +942,15 @@ export class Pool {
...
-    const loanNoteAccount = await marginAccount.withGetOrCreatePosition({
-      positionTokenMint: this.addresses.loanNoteMint,
-      instructions
-    })
-
     // If swapping on margin
-    const accountPoolPosition = marginAccount.poolPositions[this.symbol ?? ""]
+    const accountPoolPosition = marginAccount.poolPositions[this.symbol]
     if (swapAmount.gt(accountPoolPosition.depositBalance) && marginAccount.pools) {
       await marginAccount.refresh()
```
Tristyn, line 948: Is this necessary?

PR #250, opened 2022-08-17
closeAccounts and refresh position

loganjross:


libraries/ts/src/margin/pool/pool.ts
```diff
@@ -905,6 +906,7 @@ export class Pool {
       minAmountOut
     })
 
+    await marginAccount.refresh()
```
Tristyn, line 909: Why? It's refreshing before the transaction is sent not after. Maybe this should not be here

Tristyn (approved):

PR #264, opened 2022-08-26
better logging for invalid pool price during refresh position

dnut:
prints the specific pyth error message before swallowing it and replacing it with a generic one


suscd (approved):

Tristyn:
Nice

PR #265, opened 2022-08-29
JET-813: Working on fixing the shiftBy levswap issue

wilbarnes:
i am trying to update `margin-swap` so that on withdraw an additional parameter is included to determine the appropriate `TokenChange` for the withdraw

i made most of the changes i think i need to make, and got the cargo tests to pass

but the anchor tests still pass, despite making changes


programs/margin-swap/src/instructions/margin_spl_swap.rs
```diff
@@ -194,10 +198,11 @@ pub struct SwapInfo<'info> {
 
 pub fn margin_spl_swap_handler(
     ctx: Context<MarginSplSwap>,
+    change_kind: ChangeKind,
     amount_in: u64,
     minimum_amount_out: u64,
 ) -> Result<()> {
-    ctx.accounts.withdraw(amount_in)?;
+    ctx.accounts.withdraw(change_kind, amount_in)?;
     ctx.accounts.swap(amount_in, minimum_amount_out)?;
```
dnut, line 206: The `amount_in` parameter changes meaning here. When SetTo is used, it's no longer the `amount_in` quantity for a swap because for example SetTo 0 means "swap my full balance." So this line should change to two lines.

The most straightforward approach would be to copy the strategy used to determine the deposit amount at the end of this instruction. Check the balance of the transit source account and simply swap the full balance:
```suggestion
    let amount_in = token::accessor::amount(&ctx.accounts.transit_source_account)?;
    ctx.accounts.swap(amount_in, minimum_amount_out)?;
```

My only concern with this approach is that it assumes the transit account is empty to begin with, or that if there is a pre-existing balance, the user is happy to drain it for a swap. The same assumption is already built into `transit_destination_account`, so it seems OK to apply the same assumption to the source for now. Alternatively, we could check the balance before the withdrawal, and only swap the difference.

programs/margin-swap/src/instructions/margin_spl_swap.rs
```diff
@@ -194,10 +198,11 @@ pub struct SwapInfo<'info> {
 
 pub fn margin_spl_swap_handler(
     ctx: Context<MarginSplSwap>,
+    change_kind: ChangeKind,
     amount_in: u64,
```
dnut, line 202: I would probably change the name of this parameter because it actually represents the number used for withdrawal, and is not necessarily the same as the `amount_in` for the swap. Maybe `withdraw_amount` would be better since it's used for the amount parameter in the withdraw instruction.

libraries/rust/src/ix_builder/margin_swap.rs
```diff
@@ -85,6 +86,8 @@ impl MarginSwapIxBuilder {
         source_pool: &MarginPoolIxBuilder,
         destination_pool: &MarginPoolIxBuilder,
 
+        // added TokenChange here
+        change: TokenChange,
         amount_in: u64,
```
dnut, line 91: I think we can remove `amount_in` here since TokenChange replaces it with the tokens field.

dnut (changes requested):

jetjake (approved):
this looks good to me

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1313,7 +1318,7 @@ export class Pool {
       ),
       // adapterMetadata: new PublicKey("DUheebnZrHMGzEMbs9FpPFTkbmVdZnyW92CVwrYd3aGa"),
       adapterInstruction: await this.programs.marginSwap.methods
-        .marginSwap(swapAmount.lamports, minAmountOut.lamports)
+        .marginSwap(changeKind.changeKind, changeKind.value, minAmountOut.lamports)
```
Tristyn, line 1321: This may allow anchor to correctly serialize it if you were having problems
```suggestion
        .marginSwap(changeKind.changeKind.asParam(), changeKind.value, minAmountOut.lamports)
```

Tristyn (commented):
Check out my suggestions if you have trouble serializing the margin swap method

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1202,9 +1202,14 @@ export class Pool {
...
 
-    // If swapping on margin
+    // if swapping total balance or on margin (gt total balance)
+    let changeKind = PoolTokenChange.shiftBy(swapAmount)
     const accountPoolPosition = marginAccount.poolPositions[this.symbol]
-    if (swapAmount.gt(accountPoolPosition.depositBalance) && marginAccount.pools) {
+    if (
+      (swapAmount.eq(accountPoolPosition.depositBalance) || swapAmount.gt(accountPoolPosition.depositBalance)) &&
+      marginAccount.pools
+    ) {
```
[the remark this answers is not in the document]
loganjross: it doesn't exist on the TokenAmounts anymore, we tried

programs/margin-swap/src/instructions/margin_spl_swap.rs
```diff
@@ -129,10 +129,10 @@ impl<'info> MarginSplSwap<'info> {
                 self.swap_info.swap_pool.key,
                 self.swap_info.authority.key,
                 &self.margin_account.key(),
-                self.transit_source_account.key,
+                &self.transit_source_account.key(),
                 self.swap_info.vault_into.key,
                 self.swap_info.vault_from.key,
-                self.transit_destination_account.key,
+                &self.transit_destination_account.key(),
```
dnut, line 135: Why copy the pubkey? It may be a negligible difference, but I expect this to be more costly.

dnut (changes requested):

jetjake (dismissed):

dnut (approved):

PR #280, opened 2022-09-06
Adapter Collateral

dnut:
AdapterCollateral: A new type of collateral position where the position token account is owned and registered directly by a trusted adapter instead of the margin account (just like a claim).

### Specific use case
In bonds we want the following behavior (this would be useful for any order book adapter):
- open orders within bonds have collateral value in margin
- filled orders are automatically settled to an actual deposit of real money into a margin account without requiring the margin account owner to do anything
- the collateral value assigned to the margin account for the open order is decremented atomically with the settlement. this prevents the user from being credited with more collateral value than they deserve

A similar feature is also desired for staked bond tickets.

### General solution: Adapter-managed collateral as a new type of margin account position
Normally, the margin account needs to directly custody collateral tokens. This is the safest approach for fungible tokens. The token has intrinsic value, and a liquidator can easily convert that token directly into a more useful asset, like USDC, regardless of the user. This PR proposes no change to this standard behavior for deposits of fungible tokens.

Sometimes a user will have a position with its intrinsic value tied up within an adapter. This position is specific to a user and is not fungible. For example a user may have made a USDC deposit into serum to place an order on the orderbook. This order can be thought of as having some liquid value since it can be cancelled, resulting in a USDC redemption. But the order itself cannot be converted into a fungible token that is fully custodied by the margin account solely as an spl token. It has unique attributes that must be associated with only that particular order. This data must be owned and managed by the adapter program.

In this case, if we want the open order to be treated as collateral in margin, we may want to provide some channel for a trusted adapter to communicate the balance contained within that position. Since the adapter already owns the actual position, this means the real position may change without the user taking action. So to enable the adapter to communicate the real size of that position at any time, it needs to be able to communicate any changes in size to the margin program without needing the margin account to sign the transaction.

Margin expects all positions to be represented with a token account. So this change allows an adapter to own a token account that is registered as a position in the margin account, and it controls the balance of that token account to represent any arbitrary changes to the position contained within that adapter. In other words, this collateral position needs to be handled in the same way as a claim. Likewise it entails the same risks as claims, meaning we must fully trust that the adapter will safely represent a real form of collateral that can be relied upon for liquidations. It also involves the same protections adapters get with claims, in that a user may not misappropriate the tokens that are used to represent a position that is actually controlled by an adapter.


programs/margin/src/state.rs
```diff
@@ -584,6 +584,9 @@ pub enum PositionKind {
 
     /// The position contains a balance of tokens that are owed as a part of some debt.
     Claim,
+
+    /// The position contains a balance of tokens that are owed as a part of some debt.
```
suscd, line 588:
```suggestion
    /// The position contains a balance of available collateral, where the token account is managed by an adapter
```

suscd (approved):

suscd (approved):

PR #282, opened 2022-09-07
projecting risk after a margin swap

loganjross:


Tristyn (changes requested):
Please undo the changes to package-lock.json and yarn.lock. You might've ran `npm i` instead of `yarn`

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1801,6 +1810,56 @@ export class Pool {
...
+    const outputTokenLiabilityValue = outputTokenPosition.loanBalance.tokens * outputTokenPrice;
+
+    // Collateral values
+    const requiredCollateral = marginAccount.valuation.requiredCollateral.toNumber();
+    const weightedCollateral = marginAccount.valuation.weightedCollateral.toNumber();
+    const inputTokenWeight = this.depositNoteMetadata.valueModifier.toNumber();
+    const outputTokenWeight = outputToken.depositNoteMetadata.valueModifier.toNumber();
+
+    // Projected risk equation
+    const riskIndicator = (totalLiabilities + requiredCollateral - (((1 + outputRequiredCollateralFactor)/outputRequiredCollateralFactor) * Math.abs(outputTokenLiabilityValue - minAmountOutValue)) + (((1 + inputRequiredCollateralFactor)/inputRequiredCollateralFactor) * Math.abs(inputSwapValue - inputTokenAssetValue))) / (weightedCollateral + (outputTokenWeight * Math.abs(minAmountOutValue - outputTokenLiabilityValue)) - (inputTokenWeight * Math.abs(inputTokenAssetValue - inputSwapValue)))
```
eyevz, line 1853: Pretty close. The use of `abs` is not correct here. You're looking for `max(value, 0)`.

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1801,6 +1810,56 @@ export class Pool {
...
+    const outputTokenPrice = outputToken.prices.priceValue.toNumber();
+
+    // Swap values
+    const inputSwapValue = amount  * inputTokenPrice;
+    const marginAmount = amount - marginAccount.poolPositions[this.symbol].depositBalance.tokens;
+    const marginAmountValue = marginAmount * inputTokenPrice;
+    const minAmountOutValue = minAmountOut * outputTokenPrice;
+
+    // Total Liabilities
+    const totalLiabilities = marginAccount.valuation.liabilities.toNumber() + marginAmountValue;
```
eyevz, line 1836:
```suggestion
    const totalLiabilities = marginAccount.valuation.liabilities.toNumber();
```
This is just the liabilities of the margin account before the swap. The adjustment for a possible borrow is contained in the expression for `riskIndicator` below.

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1801,6 +1810,56 @@ export class Pool {
...
+
+    // Position-specific valuations
+    const inputTokenPosition = marginAccount.poolPositions[this.symbol];
+    const outputTokenPosition = marginAccount.poolPositions[outputToken.symbol];
+    const inputRequiredCollateralFactor = inputTokenPosition.loanPosition?.valueModifier.toNumber() ?? 0;
+    const inputTokenAssetValue = inputTokenPosition.depositBalance.tokens * inputTokenPrice;
+    const outputRequiredCollateralFactor = outputTokenPosition.loanPosition?.valueModifier.toNumber() ?? 0;
```
eyevz, line 1843: It would be a fatal error if the required collateral factors are missing. Setting them to zero won't save you because you divide by them below...

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1801,6 +1810,56 @@ export class Pool {
...
+    const outputTokenWeight = outputToken.depositNoteMetadata.valueModifier.toNumber();
+
+    // Projected risk equation
+    const riskIndicator = (totalLiabilities + requiredCollateral - (((1 + outputRequiredCollateralFactor)/outputRequiredCollateralFactor) * Math.abs(outputTokenLiabilityValue - minAmountOutValue)) + (((1 + inputRequiredCollateralFactor)/inputRequiredCollateralFactor) * Math.abs(inputSwapValue - inputTokenAssetValue))) / (weightedCollateral + (outputTokenWeight * Math.abs(minAmountOutValue - outputTokenLiabilityValue)) - (inputTokenWeight * Math.abs(inputTokenAssetValue - inputSwapValue)))
+
+    // TODO: add pool projections for rates
+    return {
+      riskIndicator,
+      depositRate: defaults.depositRate,
+      borrowRate: defaults.borrowRate
```
eyevz, line 1859: I suggest you return `nan` or some sentinel that indicates these values are meaningless, or modify the signature to just return the risk indicator.

After a swap, two pools have their rates affected. Even if you extracted the change in rates for the input token pool here, you'd have to separately calculate it for the output pool. We don't, or shouldn't, show the changes in pool rates caused by a swap anyway.

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1801,6 +1810,56 @@ export class Pool {
...
+    outputToken: Pool | undefined
+  ): PoolProjection {
+    const defaults = this.getDefaultPoolProjection(marginAccount);
+    if (!minAmountOut || !outputToken) {
+      return defaults;
+    }
+
+    // Prices
+    const inputTokenPrice = this.prices.priceValue.toNumber();
+    const outputTokenPrice = outputToken.prices.priceValue.toNumber();
```
eyevz, line 1827: Please confirm: "input" means token sent to the swap program, "output" means token received?

eyevz (changes requested):

libraries/ts/src/margin/pool/pool.ts
loganjross: This makes sense but would break the frontend for now, I will add a TODO though

Tristyn (dismissed):

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1799,6 +1808,63 @@ export class Pool {
     return { riskIndicator, depositRate, borrowRate }
   }
 
+  /// Projects the user's risk level after a swap.
+  projectAfterMarginSwap(
+    marginAccount: MarginAccount, 
+    amount: number,
+    minAmountOut: number | undefined,
+    outputToken: Pool | undefined,
+    setupCheck?: boolean
```
eyevz, line 1817: Nice idea to build this in here. Will need one or two modifications below to accommodate.

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1799,6 +1808,63 @@ export class Pool {
...
+    }
+
+    // Prices
+    const inputTokenPrice = this.prices.priceValue.toNumber();
+    const outputTokenPrice = outputToken.prices.priceValue.toNumber();
+
+    // Swap values
+    const inputSwapValue = amount  * inputTokenPrice;
+    const marginAmount = amount - marginAccount.poolPositions[this.symbol].depositBalance.tokens;
+    const marginAmountValue = marginAmount * inputTokenPrice;
```
eyevz, line 1831:
```suggestion
```
I think you don't need these anymore.

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1799,6 +1808,63 @@ export class Pool {
...
+
+    // Position-specific valuations
+    const inputTokenPosition = marginAccount.poolPositions[this.symbol];
+    const outputTokenPosition = marginAccount.poolPositions[outputToken.symbol];
+    if (!inputTokenPosition.loanPosition || !outputTokenPosition.loanPosition) {
+      console.error("No pool positions for margin account");
+      return defaults;
+    }
+
+    const inputRequiredCollateralFactor = inputTokenPosition.loanPosition.valueModifier.toNumber();
```
eyevz, line 1845: In setup check mode, this should be multiplied by SETUP_LEVERAGE_FRACTION.

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1799,6 +1808,63 @@ export class Pool {
...
+    const inputTokenAssetValue = inputTokenPosition.depositBalance.tokens * inputTokenPrice;
+    const outputRequiredCollateralFactor = outputTokenPosition.loanPosition.valueModifier.toNumber();
```
eyevz, line 1847: In setup check mode, this should be multiplied by SETUP_LEVERAGE_FRACTION.

eyevz:
@loganjross what branch can I look at in the UI to see how this swap projected risk indicator is used? Are you already using it to perform the setup check, as well as forecast the risk indicator?

libraries/ts/src/margin/pool/pool.ts
```diff
@@ -1856,7 +1855,9 @@ export class Pool {
 
     const inputRequiredCollateralFactor = inputTokenPosition.loanPosition.valueModifier.toNumber()
```
eyevz, line 1856: This one also needs to be scaled when in setup check mode.

eyevz (changes requested):

eyevz (approved):
Projected risk looks good. The rates are not projected but, as discussed, they don't need to be shown in this context and we can return to the TODO note later.

Tristyn (approved):

PR #304, opened 2022-09-14
Allow fetching more than 100 tokens by batching

Tristyn:
There is a limit of 100 accounts when using `getMultipleAccountInfos` which is exceeded when fetching mints, so this PR batches every unbounded call to that method.

There is no limit as far as I can see for `getTokenAccountsByOwner`


libraries/ts/src/token/associatedToken.ts
```diff
@@ -295,6 +300,29 @@ export class AssociatedToken {
     return accounts
   }
 
+  /**
+   * Fetch all the account info for multiple accounts specified by an array of public keys.
+   * This
```
Tristyn, line 305:
```suggestion
```

jetjake (approved):

PR #305, opened 2022-09-14
Register any token as a position without pool/adapter/control/metadata

dnut:
## Summary
Previously, no token could be registered as position unless a lending pool was already created for that token. Then only the deposit and loan notes could actually be used as positions and everything had to go through pools. The registration was done through the control program during pool creation. USDC could never be a position - only USDC deposit notes.

Now it is possible to register a position for any arbitrary token mint once the individual token mint is approved by the dao. There is no longer a need for a lending pool or an adapter of any kind to manage the token.

As always, any token can be owned by a margin account without the dao's approval, but it will not count as collateral until explicitly registered as a collateral token.

#### fixed term lending
The immediate need for this change is to register fixed term claims. This allows us to easily add arbitrary new position tokens for any use case without requiring any code changes.

#### margin wallet
This enables users to hold direct collateral positions on arbitrary SPL assets such as USDC or SOL without having to lend it to a lending pool. This is a big step towards using the margin account as a generic SPL token wallet. The two remaining pieces in the program to reach that goal are moving to ATAs and beefing up the token account checks in adapter_invoke to enable arbitrary token program invocations by users.

#### reduced development cost for new features
Previously we needed a special program written any time we wanted to translate different position types. For example margin-swap was needed to convert any lending pool deposit into a deposit in another lending pool. Likewise margin-swap was required for the levswap (margin trading) use case. This change makes the margin-swap program redundant. We can simply remove margin-swap, and use any swap program we want, not just orca v2, for leveraged swaps.

## Detail
- token metadata is registered in a margin pda instead of metadata program
- TokenMetadata and PositionTokenMetadata structs are no longer used. There is no longer a distinction between token and position token. now there is only `TokenMeta`. However, `underlying` is kept and optional: one token can be another's underlying (regardless of whether the underlying has its own TokenMeta)
- authority must be either the same authority as control or the control program itself
- lending pools are still created by the control program. it calls margin to register the deposit and loan notes. the underlying is not registered here any more, as it should be registered independently
- tokens may have margin specified as the adapter to allow direct pricing without an adapter, simplifying position refreshes
- tokens can optionally be priced equivalently to their underlying

## Compatibility
This changes the API for several instructions because it uses a new account in margin instead of old accounts from metadata. All the libraries have been updated to be compatible and the tests pass. We should do a little more manual testing before deployment to make sure the UI is good.

There should be no impact on existing margin accounts and their existing positions. Lending pools themselves remain valid and unchanged. However, to allow margin accounts to register positions on a lending pool they have not used before, we'll need to register the existing deposit and loan notes as valid position tokens in margin. That should be done at the same time this is deployed.

## misc thoughts
- PositionKind::NoValue has no utility as is, and we should probably repurpose it to mean "this token cannot be registered as a position." maybe it should be renamed as well


dnut:
closed in favor of sherard's similar change that came with airspaces

PR #325, opened 2022-09-22
Bonds integration webpack 5

Tristyn:
Opening this for you @davodesign84


.gitignore
```diff
@@ -1,17 +1,22 @@
 ignore/**
-**/lib
-**/dist
+lib
+.dist
```
Tristyn, line 3: Errant dot?
```suggestion
dist
```

dnut (approved):
After merge, I'd like to consider what's going on in setup.sh and determine what it's needed for and if it should be documented, added to some other process, or removed.

Also worth noting is the recommended anchor cli version in the readme being changed to 0.24.2. The programs use anchor_lang 0.25.0, and CI builds use the cli version 0.25.0. I realize the ts libs are using 0.24.2 so it's fuzzy what version to recommend. The version of the cli usually doesn't even matter except for occasional breaking changes or new features we want to use. Hopefully we can get everything on the same version soon.

PR #326, opened 2022-09-22
fixed term interest rate <> price conversions

dnut:
This implements the conversion between bond ticket prices and interest rates. There are three implementations side by side that all accomplish this goal in a different way and thus produce different values. The difference is that each one has a different interpretation of what "interest rate" means. These are:
- APY (polynomial) - annual rate that includes any effects of compounding tenors shorter than a year. For tenors longer than a year, it reflects the tenor price by compounding the APY each year until it reaches the tenor duration.
- APR (exponential) - annual uncompounded rate that can be compounded continuously over the duration of a tenor to calculate the actual yield or price of an individual tenor. (this is a nominal rate with continuous compounding used as a standard to make all tenors comparable)
- Nominal: tenor yield scaled linearly by `1y/tenor` (this is a nominal rate with a different compounding period for each tenor)

I believe that either APY or APR is a fine choice because they enable a meaningful comparison of different tenors. In other words, if daily and a monthly tenors have the same APY or APR, then you know they are equally profitable investments over any duration that involves a single deposit at the start and withdrawal at the end. On the other hand, the linear approach assumes distinct treatment of different tenors (i.e. shorter tenors involve frequent withdrawals to avoid compounding, whereas longer tenors have no such withdrawals), so if two tenors have the same linear rate that doesn't mean they are equally profitable over any time period. See the tests and scenario notes in methods.rs for more details on this topic.

For the actual api I've exposed the APR implementation. I've left all three implementations so we can easily swap out the implementation, since I don't think we've finalized on this decision yet.


libraries/ts/bonds/wasm-utils/src/orderbook/methods.rs
```diff
@@ -42,42 +47,54 @@ pub fn get_orders_from_slab(slab_bytes: &[u8]) -> Array {
...
-/// NOTE: price is returned in fixed point 32 format
+/// Interest rate is given as basis points. Tenor is in seconds.
+///
+/// NOTE: price is returned in fixed point 32 representation
 #[wasm_bindgen]
-pub fn rate_to_price(_interest_rate: f64, _duration: u64) -> u64 {
-    todo!()
+pub fn rate_to_price(interest_rate: u64, tenor: u64) -> u64 {
+    let year_proportion = Number::from(tenor) / SECONDS_PER_YEAR;
+    let rate = Number::from(interest_rate) / 10_000;
```
jetjake, line 74: 1. What is a typical interest rate that we expect? 
2. Is there an interest rate that would fall outside of the bounds of what is sane/expected?
3. is there a check somewhere for scream if this rate exceeds those bounds?

jetjake (commented):
Looks fine, just one question for my knowledge.

libraries/ts/bonds/wasm-utils/src/orderbook/methods.rs
zachanon: 1. @eyevz would have a better idea than me.
2 + 3. Almost certainly, but as we can upcast, the only case that should happen in is if an arburd basis points value or tenor gets passed in, such that it overflows. The test should be modified to fuzz within reasonable boundaires otherwise.

dnut: I just added some panics in the code for edge cases where it would produce values that make no sense, either because the type cannot represent the number or because the concept is undefined. Negative and infinite interest rates for example. I also added a few tests to ensure the code panics in those situations. Also added some tests to check the logic at acceptable edge cases, like a price of 1.

I haven't addressed the concept of typical/reasonable/sane beyond this. If someone wants to offer a loan for zero interest, or someone wants to borrow a loan for a billion percent interest, either of those seem fine here. The math logic shouldn't care about that imo. It should just make sure it provides a meaningful and correct number in response to the request. If we want to prevent people from getting into unfavorable situations then I would either totally block it in the program or discourage it explicitly in the frontend.

Tristyn (approved):

PR #330, opened 2022-09-22
JET-943 Flight logs handle for dust repay ix

rhubarbgarden:


libraries/ts/src/margin/marginClient.ts
```diff
@@ -224,7 +224,8 @@ export class MarginClient {
         // If trade action is swap, set up input amount as well
         // Get value of amount in the first parsed instruction
         if (tradeAction === "swap" && parsedIxnArray[0]) {
-          amountIn = new BN(parsedIxnArray[0].parsed.info.amount)
```
jetjake, line 227: So now, we're just changing one fixed value with another fixed value and if our program execution ordering changes at all, this will break again.  Can't we be smarter and find the transfer that happens from the source transit account into the orca account, and from the orca account to the destination account?

libraries/ts/src/margin/marginClient.ts
```diff
@@ -255,8 +256,8 @@ export class MarginClient {
                   }
                 })
               })
-              const firstTransferIxSource: string = transferIxs[0].parsed.info.source
-              const finalTransferIxSource: string = transferIxs[transferIxs.length - 1].parsed.info.source
+              const firstTransferIxSource: string = transferIxs[1].parsed.info.source
```
jetjake, line 259: Same as above.

jetjake (changes requested):
Not a scalable solution.

PR #331, opened 2022-09-22
wasm back to 10.0.3

jetjake:


Tristyn:
Being on 0.10.3 breaks windows, and 0.10.1 breaks mac. It's a mess, maybe we should fork it, use that and make a PR?

Tristyn:
I rerun checks and they pass.
The 2k changes to yarn.lock is concerning so I rebuilt the pr at #334

PR #335, opened 2022-09-26
add airspaces API

suscd:
- add permission API called airspaces
- add new config state in margin program, to be migrated from metadata in the future
- add support for registering token accounts directly owned by the margin account (for deposits, instead of having to use a pool)
- fix/workaround IDL generation issues with margin program

- does not include breaking changes to existing programs, accounts and pools do not yet respect airspaces


programs/airspace/src/instructions/airspace_create.rs
```diff
@@ -0,0 +1,65 @@
...
+    default_directives: Account<'info, DefaultDirectives>,
+
+    system_program: Program<'info, System>,
+}
+
+pub fn airspace_create_handler(
+    ctx: Context<AirspaceCreate>,
+    _seed: String,
+    is_restricted: bool,
+    authority: Pubkey,
```
dnut, line 52: why is this a parameter instead of an account?

programs/margin/src/util.rs
```diff
@@ -177,6 +186,22 @@ impl std::fmt::Debug for BitSet {
     }
 }
 
+/// Reclaim rent from an account, thereby closing it
+///
+/// This will not uninitialize the account, and should not be relied upon as a means of
+/// destroying or preventing the reuse of an acccount.
+pub fn close_account(account: &AccountInfo, receiver: &AccountInfo) -> anchor_lang::Result<()> {
```
dnut, line 193: what's wrong with AccountsClose::close?

suscd: nothing, not sure why that's not used. fixed

programs/airspace/src/instructions/airspace_create.rs
suscd: doesn't need to be an account, and technically its lower cost for the runtime

programs/margin/src/adapter.rs
```diff
@@ -287,38 +290,55 @@ fn register_position(
...
-    let metadata = metadata.log_on_error("position token metadata not found for mint")?;
+
     let token_account = token_account.log_on_error("position token mint not found")?;
     let mint = mint.log_on_error("position token account not found")?;
 
     if mint.key() != token_account.mint {
         msg!("token account has the wrong mint");
         return err!(ErrorCode::PositionNotRegisterable);
     }
-    if metadata.token_kind != TokenKind::AdapterCollateral
```
dnut, line 305: Why did you get rid of this check? Do you think it would be appropriate for an adapter to register a deposit position?

suscd: I'm not sure it _really_ matters if an adapter can do that, though the check is not removed, just moved to the `may_be_registered_or_unregistered` function on the position.

programs/margin/src/state/account/positions.rs
```diff
@@ -267,13 +288,24 @@ impl AccountPosition {
...
-                PositionKind::Claim | PositionKind::AdapterCollateral => {
-                    approvals.contains(&Approver::Adapter(self.adapter))
-                }
+        let mut authority_approved = false;
+        let mut adapter_approved = false;
+
+        for approval in approvals {
+            match approval {
+                Approver::MarginAccountAuthority => authority_approved = true,
+                Approver::Adapter(_) => adapter_approved = true,
```
dnut, line 297: Seems that the main thing you changed here was that you removed the adapter restriction. You're allowing any adapter to register a position regardless of which adapter is in charge of that token. So for example the bonds program can register a loan in pools. I don't think we should allow that kind of thing.
```suggestion
                Approver::Adapter(approving_adapter) => adapter_approved = approving_adapter == self.adapter,
```

programs/margin/src/adapter.rs
dnut: Actually I see you just moved this check into the approvals verification, which should accomplish the same thing. It's a little unintuitive to me to think of an approval as something that would prevent the execution of the thing that is being approved. But I guess it's OK.

programs/margin/src/state/config.rs
```diff
@@ -0,0 +1,162 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use anchor_lang::prelude::*;
+
+use crate::ErrorCode;
+
+/// Description of the token's usage
+#[derive(AnchorSerialize, AnchorDeserialize, Eq, PartialEq, Clone, Copy, Debug)]
+pub enum TokenKind {
```
dnut, line 24: Do we really need another enum for this? What's wrong with PositionKind?

Also I don't see the purpose of NoValue. A NoValue token is a token that should not be registered.

programs/margin/src/adapter.rs
dnut: didn't notice your response lol

programs/airspace/src/instructions/airspace_create.rs
dnut: It also means transactions cannot dedupe the pubkey.

I guess I can see how it's less burden on a validator because the data doesn't need to be accessed and the metadata doesn't need to be put into an AccountInfo. How does it actually impact our program execution though? I guess the memory limit?

I guess none of these performance concerns are very important for an instruction like this.

programs/margin/src/state/config.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// If this is `None`, then the margin program is in control of this asset, and
+    /// thus determining its price. The `oracle` field must be set to allow the margin
+    /// program to price the asset.
+    pub adapter_program: Option<Pubkey>,
+
+    /// The oracle for the token
+    ///
+    /// This only has effect in the margin program when the price for the token is not
+    /// being managed by an adapter.
+    pub oracle: Option<TokenOracle>,
```
dnut, line 82: This allows an invalid state. The oracle and adapter can both be None. The position needs to be priceable somehow. It's nice when the type system can provide compile-time guarantees about well formed data.

You could get rid of the Option here and have TokenOracle represent all valid pricing situations:
```rust
pub enum TokenOracle {
    Adapter(Pubkey),
    Pyth {
        /// The pyth address containing price information for a token.
        price: Pubkey,

        /// The pyth address with product information for a token
        product: Pubkey,
    },
}
```

programs/margin/src/state/config.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    /// The maximum staleness (seconds) that's acceptable for balances of this token
+    pub max_staleness: u64,
+}
+
+impl TokenConfig {
+    pub const SPACE: usize = 8 + 2 + std::mem::size_of::<Self>();
+
+    pub fn validate(&self) -> Result<()> {
+        if self.underlying_mint == Pubkey::default() {
```
dnut, line 101: What is the underlying mint for USDC?

programs/margin/src/state/config.rs
suscd: sure, we can clean that up now. Replacing `PositionKind` with the new `TokenKind`, and removed `NoValue`.

programs/margin/src/state/config.rs
suscd: The underlying mint for USDC is USDC. This makes it easy to search for any USDC-like token by just looking at the underlying mint.

programs/margin/src/state/config.rs
suscd: sure, combined the fields into a single enum to avoid the possibility

programs/margin/src/state/config.rs
dnut: OK. I don't feel strongly about this. I see the advantage to your approach and I'm not opposed to it. But anyway some context on my thought process...

Normally I would make this assumption: A financial instrument is a derivative if and only if it has an underlying instrument.

I could imagine wanting to search exclusively for either derivatives or primary assets, which would be easy if you could filter on the existence of the underlying field. For example, somebody may be interested in just seeing all the primary assets that can be deposited without regard for all the varieties of deposit notes may be available. I guess it's not too hard to do that with your approach. Define "self == underlying" as "no underlying" and you can filter for primary assets by excluding any that don't have that equality. Though that requires additional information about the data.

programs/airspace/src/state.rs
```diff
@@ -0,0 +1,124 @@
...
+
+declare_account_size!(GovernorId, 40);
+
+/// The set of directives are configuration for resources in an airspace, which are managed
+/// by the protocol governance. The authority for an airspace has no control over this
+/// configuration.
+#[derive(Default, AnchorDeserialize, AnchorSerialize, Debug, Clone, Copy)]
+pub struct Directives {
+    /// The fee applied to interest earned on margin pools.
+    pub margin_pool_management_fee_rate: u16,
```
dnut, line 101: It feels weird to put this in the airspaces program. I thought airspaces would be more generic.

Adapters already need to be airspace-aware before they can apply any airspace-specific behavior. So now there is this cyclic relationship between programs. Putting this data in airspaces is an assumption that margin-pools is going to behave a specific way. Why not encapsulate that logic into margin-pools? If we want a single fee rate applied to all margin pools in an airspace, why not have the program that already needs to associate pools to an airspace (margin-pools) also manage the relationship of fee-rate to airspace? I'm not even sure why we need to configure a single fee-rate per airspace. Maybe pools should allow a separate rate per pool.

I expect that we'll want to change this struct in the future which has migration implications. The complexity of this Directives struct has unbounded growth potential as we increase the number of adapters. We'll need to redeploy airspaces each time any adapter decides to modify an airspace-to-something relationship that's managed by the governor. For any changes, we'll need to test every single adapter that owns some of the data in here.

I think it would be a clean separation of concerns to let adapters deal with all the adapter-specific logic, and try to make airspaces simple and generic without requiring additional complexity that could compromise the entire margin ecosystem. It would be nice to isolate adapter-specific risks into their own little sandboxes as much as possible.

programs/airspace/src/instructions/set_governor.rs
```diff
@@ -0,0 +1,65 @@
...
+            // In testing environments, governor can be set by first caller
+            governor_id.governor = governor.key();
+        } else {
+            // In production/mainnet, governor has hardcoded default
+            governor_id.governor = GOVERNOR_DEFAULT;
+        }
+    }
+
+    // Verify the signer is actually current governor
+    if governor_id.governor != governor.key() {
```
dnut, line 57: This needs to be executed by the default governor before the first action they take. Seems fine I guess but it's an unnecessary restriction. One more step to push through governance.

An alternative approach could make initialization permissionless as long as it's not trying to set an unexpected governor. Whoever deploys the program can also initialize the governor. It's a more complete deployment.
```suggestion
    if governor_id.governor != governor.key() && new_governor != governor_id.governor {
```

programs/airspace/src/instructions/set_governor.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Accounts)]
+pub struct SetGovernor<'info> {
+    #[account(mut)]
+    payer: Signer<'info>,
+
+    /// The current governor
+    governor: Signer<'info>,
+
+    /// The governer identity account
+    #[account(init_if_needed,
```
dnut, line 31: I wonder if this is the best case for init_if_needed since there is initialization specific logic in here. Maybe the behavior/security would be more obvious if it's split into two instructions.

suscd: Split into another initialization instruction, kept the `init_if_needed` to allow passing even if it's already initialized.

programs/airspace/src/instructions/set_governor.rs
suscd: Sure, splitting the initialization does seem cleaner, so added an extra `create_governor_id` to do that, which doesn't require the governor to sign.

programs/airspace/src/state.rs
suscd: I don't really disagree that its weird, it's just convenient from the perspective of having a single place for setting defaults. Otherwise every adapter now needs to have some extra instructions/state to manage global defaults. An alternative might be to just hardcode the defaults in the programs for now.

dnut (approved):

PR #340, opened 2022-09-26
Try installing rust

jetjake:


dnut:
I think rust should just be a prerequisite. I don't like the idea of an app's build process automatically installing software into my system.

jetjake:
@dnut its for cloudflare pages only and I named it as such.  Which sucks but its better than committing the dist folders for now.  A better solution is building the library artifacts and pushing them to an artifact storage as part of CI and then having cloudflare pull those down, but not sure cloudflare pages supports that from what I researched today.  I am just making a marginally better change in lieu of a shittier one

jetjake:
Also, @dnut apologies, the PR should have been a draft.  That's my fault.

dnut:
Makes sense. Seems fine as a new command only used by cloudflare. I just heard the alarm bells ringing because I thought this might get merged any second, meanwhile `yarn build` is a command people are running all the time just to build the code.

PR #356, opened 2022-09-30
WASM Math tests

zachanon:
Add tests to allow confidence in the performance of the wasm mathematical methods.

### DIFF
--- a/libraries/ts/bonds/wasm-utils/src/orderbook/methods.rs
+++ b/libraries/ts/bonds/wasm-utils/src/orderbook/methods.rs
@@ -51,14 +51,15 @@ pub fn order_id_to_string(order_id: &[u8]) -> String {
 /// Given a base quanity and fixed-point 32 price value, calculate the quote
 #[wasm_bindgen]
 pub fn base_to_quote(base: u64, price: u64) -> u64 {
-    let quote = Fp32::upcast_fp32(price) * base;
-    quote.as_decimal_u64().unwrap()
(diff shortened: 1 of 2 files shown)


libraries/ts/bonds/wasm-utils/src/orderbook/methods.rs
```diff
@@ -142,3 +143,33 @@ fn wasm_uses_tested_implementation() {
         }
     }
 }
+
+/// Utilizes the same numbers as in the integrated test
+#[test]
+fn integrated_sanity_test_rate_price_covnersion() {
```
zachanon, line 149: @dnut @eyevz  We have some slippage for these values here. Not sure how we want to handle this. I suppose in theory it should be possible to have lossless conversion, given a large enough upcast. Not sure on how complex such a feat is in reality, and the overall desirability.

libraries/ts/bonds/wasm-utils/src/orderbook/methods.rs
```diff
@@ -142,3 +143,33 @@ fn wasm_uses_tested_implementation() {
...
+    let base = 1_200_000;
+    let quote = 1_000_000;
+
+    let price = calculate_implied_price(base, quote);
+
+    let derived_base = quote_to_base(quote, price);
+    let derived_quote = base_to_quote(base, price);
+
+    assert_eq!(base, derived_base);
+    assert_eq!(quote - 1, derived_quote);
```
zachanon, line 174: decided to manually account for rounding here. Is there a better way you can think of for this particular test?

PR #359, opened 2022-10-03
Conditionally render E2E wallet

rhubarbgarden:
Get value of `jetAppCluster` from localStorage and add E2E wallet to wallets array based on if mainnet or devnet.

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -22,6 +22,8 @@ const FixedLendView = lazy(() => import('./views/FixedLendView'));
 const FixedBorrowView = lazy(() => import('./views/FixedBorrowView'));
 
 export function App(): JSX.Element {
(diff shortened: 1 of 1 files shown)


qiqicodes:
When set to the "devnet" on settings, the "E2E wallet" doesn't appear on the wallet modal for the preview.
Screen Shot 2022-10-03 at 9 50 59 AM

The change does appear locally.
Screen Shot 2022-09-28 at 3 28 02 PM

jetjake (approved):

jetjake (changes requested):
what about localnet?

rhubarbgarden:
@jetjake We don't use localnet at all in the UI when running E2E tests, I don't think we would need to make the E2E wallet available to localnet

jetjake (approved):

PR #361, opened 2022-10-03
Jet 862 liquidation modal

qiqicodes:
- [x] updates the liquidation modal with one more user aciton.
   - try to end liquidation
   
<details>
<summary>Click to See the Current Liquidation Modal</summary>

</details>

<details>
<summary>Click to See the Updated Liquidation Modal</summary>

</details>


suscd:
Should we only show/enable the extra button for ending liquidation when the timeout has actually passed (and thus, it's actually possible for a user to do it)?

davodesign84:
Is there any reason why we require them to perform two separate actions? First transfer or deposit, and then try to end liquidation? Why a deposit or transfer on an account that is undergoing liquidation doesn't stop the liquidation automatically? Also why are we making the liquidation async? That gives space to all these weird scenarios. The flow that would make the most sense to me is to provide a warning if you are close to liquidation, and then liquidate immediately when the thresholds are reached, plus showing a modal if a liquidation happened on your account.

suscd:
I agree with @davodesign84 in the sense that we should only be showing users a warning when they are near or at the liquidation threshold. The fact that you can do a "end liquidation" seems like an implementation detail that shouldn't be exposed to users at all. Even in that edge case, depositing funds is still possible (for any existing position, though perhaps we could consider relaxing the limitation of creating new positions).

qiqicodes:
Decision:
* remove "under liquidation" modal to make ux more pleasant 
* bundle `end liquidation` instruction when necessary

PR #363, opened 2022-10-04
updates svg sizes to show on app

qiqicodes:
updates svg sizes to show on app


rhubarbgarden:
Wallet icons also need to be exported as SVGs.

jetjake (approved):

PR #364, opened 2022-10-04
Flightlogs blackbox

rhubarbgarden:
Initial flight logs setup from blackbox API endpoint


jetjake (approved):
Let's refactor this once devnet is up.

PR #365, opened 2022-10-04
Fixes USDConversionRates bug

qiqicodes:
- [x] Fixes USDConversionRates bug with @rhubarbgarden 
- [x] Updates modal `visible` property to `open`


app/src/components/SwapsView/SwapsGraph.tsx
```diff
@@ -211,7 +211,7 @@ export function SwapsGraph(): JSX.Element {
         labels: {
           padding: 0,
           formatter: (value: number) => {
-            return currencyAbbrev(value, false, 2);
+            return currencyAbbrev(value, false, currentPool?.precision ?? 2);
```
davodesign84, line 214: I personally prefer to use `||` instead of `??`. The difference is that `||` will return the right hand value even if the checked variable is `0` or `''` or any other falsey value, while `??` only returns the right hand value if the variable is either `null` or `undefined`. But in this case is a matter of preference unless we think that 0 decimals is a legit number.

davodesign84 (dismissed):

PR #367, opened 2022-10-05
Autopublish TS lib version on master

rhubarbgarden:
- Add `semantic release` and `commitizen`
- Add tslib publishing workflow

### DIFF
--- a/.github/workflows/tslib.yaml
+++ b/.github/workflows/tslib.yaml
@@ -11,8 +11,6 @@ on:
 jobs:
     runs-on: ubuntu-latest
 
-
-
     publish:
       runs-on: ubuntu-latest
       steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/tslib.yaml
```diff
@@ -0,0 +1,25 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Semantic Release
```
qiqicodes, line 21: semantic release would be a great addition to our existing doc release flow. 

perhaps we can add this to the existing `docs.yaml`
https://github.com/jet-lab/jet-v2/blob/c534d3f1f8ae3c8e1bd62c0e6ff1e8ceb3adace3/.github/workflows/docs.yaml#L50

PR #373, opened 2022-10-08
margin: check token account states after invoke

suscd:
Introduce a new safety check for token accounts owned by a margin account, preventing any adapter from altering any aspect of the token account other than its balance.

### DIFF
--- a/programs/margin/src/adapter.rs
+++ b/programs/margin/src/adapter.rs
@@ -19,7 +19,7 @@ use std::{collections::BTreeMap, convert::TryInto};
 
 use crate::{
     events::{PositionClosed, PositionEvent, PositionRegistered, PositionTouched},
-    util::{log_on_error, Require},
+    util::{log_on_error, Require, TokenAccountObserver},
     AccountPositionKey, AdapterPositionFlags, Approver, ErrorCode, MarginAccount, SignerSeeds,
     TokenConfig,
 };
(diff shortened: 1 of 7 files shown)


dnut:
how does this impact compute usage?

dnut:
What if a token account was initialized with arbitrary state before authority was given to the margin account? This is just going to preserve that state without making it safer. Yes, this will preserve the safety of token accounts that the margin program knows it initialized to a good state, but it also arbitrarily impacts other token accounts which don't need to be immutable.

Doesn't this block some use cases? What if a user wants to use their margin account to interact with some program that issues its own tokens and mutates other fields of the token account. The user is not interested in registering it as collateral. They just want to use that program, and that program uses this behavior internally. I see no reason why that shouldn't be allowed. Users should be allowed to use the margin account like a normal wallet with all the same features they normally have access to. It's just that if they want the special new feature we offer - treating some token as collateral - only then do we need to restrict that account to ensure custody is maintained within margin.

We know the exact bytes that should be required in collateral token accounts. No delegate, owner is the margin account, etc. Why don't we simply define the exact state expected in the token account and require that the data is equal to that for any token account that is registered as a collateral position. We can have a single function defined in one place that defines what a safe token account looks like, and reuse it anywhere that we want this guarantee, for example when registering a position.

We can construct the expected state using constants. It would probably be less expensive than the current approach, especially since we only need to apply it to token accounts that are registered as collateral positions.

PR #382, opened 2022-10-14
Add additional period for term deposits

zachanon:
Allow loan order maturation to exceed borrow duration


programs/bonds/src/control/state.rs
```diff
@@ -49,6 +49,8 @@ pub struct BondManager {
     pub(crate) _reserved: [u8; 28],
     /// Units added to the initial stake timestamp to determine claim maturity
     pub duration: i64,
+    /// Number of slots added to initial strike timestamp to determine loan maturity
```
dnut, line 52:
```suggestion
    /// Number of seconds added to initial strike timestamp to determine loan maturity
```

programs/bonds/src/control/instructions/initialize_bond_manager.rs
```diff
@@ -17,6 +17,8 @@ pub struct InitializeBondManagerParams {
     pub seed: [u8; 32],
     /// Units added to the initial stake timestamp to determine claim maturity
     pub duration: i64,
+    /// Number of slots added to initial strike timestamp to determine loan maturity
```
dnut, line 20:
```suggestion
    /// Number of seconds added to initial strike timestamp to determine loan maturity
```

programs/bonds/src/control/state.rs
```diff
@@ -49,6 +49,8 @@ pub struct BondManager {
...
+    pub deposit_duration: i64,
```
dnut, line 53: This name is a little confusing to me. I interpret it as meaning it represents the entire duration of any deposit. But it's actually only the extra delay added on to delay the redemption of staked tickets. I would probably pick some other name, something like `(ticket)?_(redemption|unstake)_(delay|latency|lag)`

dnut (changes requested):
Looks good, just some minor wording nitpicks.

programs/bonds/src/control/state.rs
dnut: Another thought - should we always have a guarantee that both durations are positive, and the lender duration is always at least as long? Maybe it's worth putting some kind of checks for that. For example using u64 instead of i64.

zachanon: I like the idea of explicitly defining the two durations. Opted not to include a check, these values are staged at initialization and then should not reasonably be changed

PR #385, opened 2022-10-17
fixes grey screen bug caused by bn comparison

qiqicodes:


davodesign84 (approved):

davodesign84:
Looks good :) Maybe add a comment on why we added the try catch, otherwise we won't remember that's an issue with that  bn function

PR #391, opened 2022-10-18
JET-1035 show how much SOL users need to do txn

KevinRafaelConnor:
Tell user how much SOL they need in their wallet if the UI is not allowing txns due to too little SOL in wallet

### DIFF
--- a/app/src/state/settings/localization/languages/Jet_UI_EN.json
+++ b/app/src/state/settings/localization/languages/Jet_UI_EN.json
@@ -60,7 +60,7 @@
       "pastTense": "Deposited",
       "disabledMessages": {
         "noBalanceForDeposit": "You have no {{ASSET}} in your wallet to deposit.",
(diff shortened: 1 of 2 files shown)


app/src/state/settings/localization/languages/Jet_UI_EN.json
```diff
@@ -60,7 +60,7 @@
       "pastTense": "Deposited",
       "disabledMessages": {
         "noBalanceForDeposit": "You have no {{ASSET}} in your wallet to deposit.",
-        "notEnoughSolForFees": "You don't have enough SOL in your wallet to cover transaction fees."
+        "notEnoughSolForFees": "`Please make sure you have a buffer of at least {{BUFFER}} SOL in your wallet to cover potential transaction fees.`)"
```
qiqicodes, line 63: we can remove the "`" and "`)".

qiqicodes (approved):
LGTM

PR #396, opened 2022-10-19
removes apricot protocol from interest radar

qiqicodes:


KevinRafaelConnor:
I was trying this approach earlier but Apricot continued to show up on my local testing of removing this.  I still see it on your branch URL too. 

I started looking at jet-api for removing Apricot on that end as well.

We can talk about it in our sync in 45 min and then go from there

KevinRafaelConnor (approved):
LGTM

PR #403, opened 2022-10-24
Constrain orders for safety

zachanon:
Interest should always be positive non-zero


eyevz:
Looks good @zachanon. Could you please add a `max_price` parameter per market and also check that `price <= max_price`?

This max price should correspond to, say, 50% annualised simple rate.

zachanon:
> Looks good @zachanon. Could you please add a `max_price` parameter per market and also check that `price <= max_price`?
> 
> This max price should correspond to, say, 50% annualised simple rate.

I agree on constraining the other bound, but are you sure on 50%? That sounds rather low for how volatile this space tends to be.

eyevz:
> ... are you sure on 50%? That sounds rather low for how volatile this space tends to be.

As discussed elsewhere, this should be a setting for individual markets. We'll watch prices and adjust if it makes sense, but I can't think of any situations for the current and foreseeable token set that would lead to annualised interest rates above 50%.

PR #410, opened 2022-10-26
Fix `ConsumeEvents` instruction

zachanon:
Currently `ConsumeEvents` fails when running on a solana validator due to a stack overflow. This moves some accounts around to try and push them onto the heap and prevent the overflow


zachanon:
Moved a bit from master, remade here to avoid dealing with rebase conflicts

https://github.com/jet-lab/jet-v2/pull/444

PR #411, opened 2022-10-26
Update events for lend and borrow order

nevi-me:


programs/bonds/src/orderbook/events.rs
```diff
@@ -13,6 +13,9 @@ pub struct LendOrder {
     pub bond_market: Pubkey,
     pub lender: Pubkey,
     pub order_summary: OrderSummary,
+    pub auto_stake: bool,
```
dnut, line 16: FYI, auto_stake is a parameter for *all* orders, not just for lending. I saw a bug once where the accounts parser in the event consumer saw this flag enabled for a borrow and assumed there was supposed to be an extra SplitTicket account included in the instruction, which caused a failure to consume events. So auto_stake is a potential avenue to debug issues with borrow orders.

My fix for that issue was to only set the AUTO_STAKE flag in the order callback for lend orders, and otherwise discard the auto_stake input. So it should be irrelevant for borrows unless this code is changed or I made a mistake. I guess something to be aware of is that the data used by the aaob may or may not be the same as the OrderParams passed into the instruction.

I don't know this has any implications for event publishing, just thought it was notable.

dnut (commented):
Why don't you add this data to the other order instructions? There are five instructions that place an order. margin_lend_order is likely to replace lend_order in the frontend as the standard way to lend.

suscd (approved):

PR #414, opened 2022-10-27
fix repay amounts

davodesign84:


eyevz (approved):
I did not test this myself, but the changes to the code make sense.

PR #417, opened 2022-10-31
Fix `SplitTicket` value calculation

zachanon:
Owning to a misunderstanding of the values in the `OrderSummary`, we were using the wrong values to calculate principal and interest


zachanon:
This particular issue was fixed in a different PR. Closing

PR #420, opened 2022-10-31
Investigate logRocket session

qiqicodes:


davodesign84:
The issue here might be that Cloudflare pages is not supporting env variables. They are set in the project but don't show up in the actual built page. Our production endpoint seems to be working through whitelisting.

PR #424, opened 2022-11-02
add loading and error state when connecting or switching network

davodesign84:
Small PR to add loading screens and show an error screen if connection to the RPC node hard fails. This situation would otherwise cause a grey screen of death for users that select the localnet option in settings when on production builds. The only way to recover from that is to clear the local storage currently. This PR would simply show an error message instead.


suscd:
If I run this locally, it works as described. If I use the cloudflare build for it, it doesn't seem to work though as a message shows briefly before switching to a blank grey screen. Any thoughts on why the behavior is different?

davodesign84:
> If I run this locally, it works as described. If I use the cloudflare build for it, it doesn't seem to work though as a message shows briefly before switching to a blank grey screen. Any thoughts on why the behavior is different?

Not fully sure of why the different behaviour but I've found a few things I missed. Hopefully will be fixed shortly.

PR #430, opened 2022-11-03
fixed term origination fee for borrowers

dnut:
Applies a small fee to fills for borrowers that create new debt. Lenders and ticket sellers will not experience fees directly. The fee is assessed as soon as the loan is issued by increasing the size of the loan and collecting a small portion of it to the protocol. The lender funds it up front (to them it just looks like a larger loan) and the borrower pays it back at the end of the term (to them it looks like they're paying slightly more interest on the same loan).

Adds origination_fee field to manager account, which is a u64 measured in units of % of bps (1/1,000,000). Scales all borrow orders by 1+fee so the actual order is slightly larger than the user requested. For every fill of a borrow order, the amount disbursed to the borrower is scaled back down by 1/(1+fee). The idea of this back-and-forth conversion is to ensure that the user get the loan that they requested, and the fee is thrown on top of that, not that they get some deduction from the loan size that they asked for.

There is a possibility that the loan amount will deviate from what they requested due to rounding error. If the entire order is filled in a single match, the user will get exactly what they requested. For each additional fill, it can lose up to 1u64 of precision. So if a borrow order has many fills, the imprecision can accumulate. But it's not an avenue for exploit because we're always making sure that enough fees are collected. It's just that the actual loans that are issued may be slightly off from what was requested, and slightly more fees than anticipated may have been collected due to a large number of fills. So you can interpret the fee as origination_fee * amount_borrowed + ~1/2u64 per fill.

The fee calculation rounds up the fee, which guarantees that a fee is assessed no matter what. If you attempt to borrow 1u64, then you'll end up with a loan disbursed of 1u64 plus 1u64 in fees, so you'll owe at least 2u64. If you attempt to borrow 1 USDC, you will pay at least 0.000001 USDC in fees per fill since that is the smallest measurable unit of USDC.

Fees are collected by keeping the money in the vault instead of disbursing it in a loan to the borrower. The amount that was held back from the borrower is added to a new field called collected_fees in the manager account. The manager account is now made mutable in the consume_events instruction in order to increment this number. Anyone can execute a new instruction WithdrawFees that sends the total balance of collected_fees from the vault to a pre-defined fee_destination token account, and zeros collected_fees. fee_destination is another new field in the manager account, which currently can only be set when the market is initialized.

misc worth a second look:
- In the clients, the fee_destination for a market is set to the authority's ATA for that token by default if they do not specify any account. Only the clients take this shortcut; the program itself requires an account to be explicitly specified.
- I have disabled preflight checks in the ctl cli since it has false positives when there are dependencies between transactions.


libraries/rust/margin/src/bonds/ix_builder.rs
```diff
@@ -762,10 +791,10 @@ impl BondsIxBuilder {
     }
 
     pub fn pause_ticket_redemption(&self) -> Result<Instruction> {
-        self.modify_manager([true as u8].into(), 8 + 32 * 13 + 2)
+        self.modify_manager([true as u8].into(), 8 + 32 * 14 + 2)
     }
     pub fn resume_ticket_redemption(&self) -> Result<Instruction> {
-        self.modify_manager([false as u8].into(), 8 + 32 * 13 + 2)
+        self.modify_manager([false as u8].into(), 8 + 32 * 14 + 2)
```
eyevz, line 797: Can't be having those unlucky offsets ;)

programs/bonds/src/control/instructions/withdraw_fees.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use crate::{bond_token_manager::BondTokenManager, control::state::BondManager};
+
+#[derive(Accounts, BondTokenManager)]
+pub struct WithdrawFees<'info> {
+    #[account(mut)]
+    pub bond_manager: AccountLoader<'info, BondManager>,
+
+    #[account(mut)]
+    pub fee_destination: AccountInfo<'info>,
```
eyevz, line 13: Is the fee destination checked against the what's been set on the bond manager? I couldn't figure out where that happens.

programs/bonds/src/margin/origination_fee.rs
```diff
@@ -0,0 +1,198 @@
...
+
+/// Based on some quote amount that was filled from an order that includes a
+/// fee, returns the size of the loan disbursement after the fee is deducted.
+pub fn loan_to_disburse(filled_quote: u64, origination_fee: u64) -> u64 {
+    if filled_quote == 0 {
+        0
+    } else if origination_fee == 0 {
+        filled_quote
+    } else if filled_quote < u64::MAX / FEE_UNIT {
+        FEE_UNIT * filled_quote / (FEE_UNIT + origination_fee)
```
eyevz, line 30: What guarantees that this is evaluated as `(u * q) / (u + f)` rather than `u * (q / (u = f))`? Is that a Rust spec, is it guaranteed for all compilers and targets?

programs/bonds/src/margin/origination_fee.rs
```diff
@@ -0,0 +1,198 @@
+const FEE_UNIT: u64 = 1_000_000;
```
eyevz, line 1: "centibeeps"? "micrumbs"? :)

programs/bonds/src/margin/origination_fee.rs
dnut: Rust guarantees left-to-right associativity of arithmetic operators with equal precedence.
https://doc.rust-lang.org/reference/expressions.html#expression-precedence

programs/bonds/src/control/instructions/withdraw_fees.rs
dnut: lol oops where are all the checks that are supposed to be on this instruction, that's not the only one missing

eyevz (changes requested):
Some questions / comments attached.

eyevz (dismissed):
Your latest changes address the main concern I had.

programs/bonds/src/margin/origination_fee.rs
```diff
@@ -0,0 +1,198 @@
...
+/// adds an origination fee.
+/// - origination_fee: scaled by FEE_RATIO
+pub fn borrow_order_qty(requested: u64, origination_fee: u64) -> u64 {
+    if requested == 0 {
+        0
+    } else if origination_fee == 0 {
+        requested
+    } else {
+        let requested = requested as u128;
+        let quote = 1 + requested + (requested * origination_fee as u128) / FEE_UNIT as u128;
```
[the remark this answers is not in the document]
dnut: It's certainly doable but it's kind of arbitrary. It would be very easy to throw in a const right here. Configuring per market is a little more complicated just because you have to deal with one more parameter in several places, but it's not a big deal.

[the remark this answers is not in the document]
dnut: > - user gets what they asked for
> - protocol gets implied origination fee plus remaining base fee?

Actually these two points are in conflict each other. The base fee would be a constant per-fill regardless of size. This *decreases* the total amount borrowed for larger numbers of fills. If the priority is to ensure users get what they requested, you'd want to simply increase the order size without assessing a base fee.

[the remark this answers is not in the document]
eyevz: I suppose I had in mind that the base fee is the part that eats the rounding errors, instead of the origination fee or requested loan, but I get that it's hard to attach metadata to an order that can be tracked across fills.

But overall I think what you've implemented here ticks all the boxes. Let's return to questions of rounding if necessary after observing this beast in the wild.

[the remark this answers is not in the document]
dnut: The idea is nice, but it puts the cart before the horse. You have to know how many fills are going to happen in the future to decide how much of a current fill should go to the user vs the base fee. Imagine an order is 50% filled by an order with 50% remaining on the book. How much do we deduct for the base fee now? 0? 50% of the base fee amount? 10%?

If we collect anything on the first fill, you need to know how many fills there will be in the future to figure out how much of the base fee to collect now. If you guess wrong, you defeat the intention of using it as a buffer because you run the risk of overcharging them.

So you can only be sure that it works properly if you collect later. That means you wait until the order is fully removed off the book, and assess the base fee during either the last fill or the order cancel. Just give them a little less money in the last fill - exactly enough ensure they get what they requested. This requires some additional state to be persisted for the order which introduces complexity that may not be worth it. Furthermore there is always the possibility that they cancel the order, in which case we have to decide: are they allowed to bypass the fee via cancellation, or do we create a new obligation for negligible amounts of money every time a partially filled order is cancelled?

Using a base fee to ensure they get exactly what they asked for seems possible but I don't think it's worth the complexity. Padding the order size by some constant will ensure they get *at least* what was requested, but possibly more, and it is much simpler to implement.

suscd (approved):

PR #438, opened 2022-11-09
Estimate order outcome

eyevz:


libraries/ts/bonds/wasm-utils/src/orderbook/types.rs
```diff
@@ -33,3 +33,28 @@ pub struct OrderAmount {
     /// fixed-point 32 limit price value
     pub price: u64,
 }
+
+#[wasm_bindgen(module = "/src/cast.js")]
```
zachanon, line 37: this is interesting, is there a doc somewhere I can read on what exactly this is doing?

PR #442, opened 2022-11-10
Constrain order prices for safety

zachanon:
Limit prices on `OrderParams` should not be above one, representing a negative interest rate. Nor should they be below a market configurable `minimum_price`, which represents a maximum interest rate.


dnut (approved):
ping me when you get a green build synced with master and I'll re-approve

zachanon:
Stale

PR #445, opened 2022-11-11
JET-1065: Add event for borrow fills

nevi-me:
When an obligation is created on a market borrow, we do not emit an event for this. I've added it so we can get the borrow fill and the obligation.

I've also updated the IDL with the new event.

### DIFF
--- a/programs/bonds/src/margin/instructions/margin_borrow_order.rs
+++ b/programs/bonds/src/margin/instructions/margin_borrow_order.rs
@@ -117,7 +117,7 @@ pub fn handler(ctx: Context<MarginBorrowOrder>, params: OrderParams, seed: Vec<u
             maturation_timestamp,
             quote_filled,
             base_filled,
-            flags: ObligationFlags::default()
+            flags: obligation.flags
         });
     }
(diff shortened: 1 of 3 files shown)


programs/bonds/src/margin/instructions/margin_borrow_order.rs
```diff
@@ -94,16 +95,30 @@ pub fn handler(ctx: Context<MarginBorrowOrder>, params: OrderParams, seed: Vec<u
...
+        emit!(ObligationCreated {
+            obligation: obligation.key(),
+            authority: ctx.accounts.margin_account.key(),
+            order_id: order_summary.summary().posted_order_id,
+            sequence_number,
+            bond_manager: bond_manager.key(),
+            maturation_timestamp,
+            quote_filled,
+            base_filled,
+            flags: ObligationFlags::default()
```
suscd, line 120: this should read back the flags from `obligation.flags`  instead of assuming default, otherwise seems easy to forget to update if there's a change to the flags value

```suggestion
            flags: obligation.flags
```

suscd (approved):

PR #446, opened 2022-11-13
Avoid unnecessary token checks during margin_repay

eyevz:
`margin_repay` currently, unnecessarily, checks that the pool contains a number of tokens equal to the value of deposit notes being burned. The check is unnecessary because no underlying tokens are involved in this operation.

This is preventing the liquidation of margin accounts with deposit and loan notes in pools with maxed out utilisation.

This PR adds a method `margin_repay` to the `MarginAccount` that does not perform the unnecessary check on the vault balance, and uses that method instead of `withdraw` and `repay` in the `margin_repay` instruction.

### DIFF
--- a/programs/margin-pool/src/instructions/margin_repay.rs
+++ b/programs/margin-pool/src/instructions/margin_repay.rs
@@ -22,7 +22,6 @@ use anchor_spl::token::{self, Burn, Token, TokenAccount};
 
 use jet_margin::MarginAccount;
 
-use crate::instructions::withdraw;
 use crate::{events, state::*, ChangeKind, TokenChange};
 use crate::{Amount, ErrorCode};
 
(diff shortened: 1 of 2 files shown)


davodesign84 (dismissed):

eyevz:
`yarn test` is working locally. Not sure what to make of failing anchor test here.

davodesign84:
> `yarn test` is working locally. Not sure what to make of failing anchor test here.

This is very odd, seems like it's running tests that are commented out. I'd say go ahead and delete the whole `tests/integration/examples` folder altogether. They don't use localnet and are redundant in the first place

davodesign84:
Even more odd, cloudflare build failed with the `build:cloudflare` not found in package.json, but it's there 🤯

davodesign84:
> Even more odd, cloudflare build failed with the `build:cloudflare` not found in package.json, but it's there 🤯

Oh wait, this is against `mainnet` branch, which is 100+ commits behind master??

suscd (approved):

PR #453, opened 2022-11-15
Fix bitcoin and market txs

davodesign84:
Fixed issues that block our CI E2E 
Also fixed token representation on the accounts table https://jetprotocol.atlassian.net/browse/JET-1034


qiqicodes (approved):
LGTM

PR #456, opened 2022-11-15
add administrative instructions for transferring deposits and loans

suscd:
These instructions allow the engineering team to manually fix accounts with bad assets by transferring the positions to a controllable account.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2239,6 +2239,7 @@ dependencies = [
  "pyth-sdk-solana 0.4.2",
  "serde",
  "serde_test",
+ "solana-program",
  "static_assertions",
 ]
 
@@ -2256,6 +2257,7 @@ dependencies = [
  "pyth-sdk-solana 0.4.2",
  "serde",
  "serde_test",
+ "solana-program",
  "static_assertions",
 ]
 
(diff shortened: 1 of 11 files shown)


dnut (approved):
Ideally I would like to see this use case served without relying on administrator/governance privileges but in our current situation it seems like the most straightforward way to move forward.

It's important to execute these in the proper order. If loans are moved before collateral, then the original owner might look for a way to withdraw their collateral. SPL governance does not restrict the executor or the order of execution of instructions within a proposal, so an attentive margin account owner might be able to exploit this. I think the safest approach would be to use two proposals. The first one moves the collateral, then when that is executed, create another proposal to move the claim. This would also make it easier to sell off the collateral before using it to repay the claim.

just a few notes as a reminder of what we already discussed and can look into later:
- ideally this could be done by a liquidator, but the discrepancy between equity vs collateral value is a blocker for the current scenario. maybe there's a way to make instructions like this executable by the liquidator
- either of the new instructions could be authorized by a margin account owner instead of the administrator (source for collateral, target for claim), though when there is bad debt we do need some override to move the collateral. but even in our current scenario we don't really need administrator privileges to move the claim. it would just require the person who is doing the repaying to approve a transaction that we don't have a UI for, which is inconvenient. i don't love the idea of using administrator permissions for convenience but i won't stand in the way either since it's the most straightforward solution.
- from a technical perspective it would be easier for various reasons if we could simply transfer the margin account ownership, but that doesn't work with the frontend due to the assumption of it being seeded by the owner
- more checks would be needed on these instructions if we were going to allow it to be executed by anyone other than governance. as is, it's probably fine. these checks would look at account health somehow depending on the situation, and make sure the token accounts are actually what we expect

PR #460, opened 2022-11-16
adding strikethrough and red text to unavailable pools

KevinRafaelConnor:


app/src/components/PoolsView/PoolsTable/PoolRow.tsx
```diff
@@ -1,10 +1,12 @@
-import { useRecoilState } from 'recoil';
+import { useRecoilState, useSetRecoilState, useResetRecoilState, useRecoilValue } from 'recoil';
```
qiqicodes, line 1: we can leave `useSetRecoilState, useResetRecoilState` out, if we are not using this in the component

PR #478, opened 2022-11-23
Adding req. coll. factor to pool info panel + hover definition

KevinRafaelConnor:
@eyevz Whenever is convenient, can you please check this video I made of an update for this ticket, and tell me if you would like anything displayed differently? Specifically:

1) The way we display the required collateral factor (as a constant made sense to me from your docs, but since there are no examples elsewhere I wasn’t positive)
2) The definition I wrote (seen in the video when I hover over the text)

https://user-images.githubusercontent.com/89974125/203619610-255aea00-88e6-44c7-b74f-9043960f1070.mov

### DIFF
--- a/app/src/components/PoolsView/PoolDetail/PoolDetail.tsx
+++ b/app/src/components/PoolsView/PoolDetail/PoolDetail.tsx
@@ -105,22 +105,6 @@ export function PoolDetail(): JSX.Element {
     return render;
   }
 
-  const linkText = "docs";
-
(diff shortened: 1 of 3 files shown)


eyevz:
Thanks @KevinRafaelConnor, looks good. Maybe a link to https://docs.jetprotocol.io/jet-protocol/protocol/margin-accounts#accounting for more info would be a nice addition to the tooltip?

eyevz (commented):
I think a link to the accounting docs would be a good addition to the tooltip.

davodesign84 (approved):

PR #487, opened 2022-12-01
Dx improvements

davodesign84:
This PR includes a variety of improvements on our front-end codebase, most of the inspired by this list https://jetprotocol.atlassian.net/wiki/spaces/JC/pages/edit-v2/69828609

It achieves the following:
- Reduce compile time (from ~65s to ~16s on my machine)
- Reduce bundle size (from ~3.8mb to ~3mb)
- Furthermore enables Lazy loading of views, so the bundle is not a single 3mb chunk, but you only load the necessary code for the view your are on
- Removes anchor tests as agreed (less computation on CI / duplicated efforts)
- Margin and Bond library are now full es module thanks to the step above, which in turn..
- Allows proper hot-reload of the app if edits happen to margin or fixed-term libraries
- Removes the need for the npm version of wasm-pack which created few issues when checking out new branches (no need to rerun `yarn add -D wasm-pack` in the fixed term lib after checking out a new branch with dependency changes. It now only uses the pure cargo version of wasm-pack
- Removes unused Serum markets code (can always go back to the history when / if needed)
- Removes .less code from the bundle. It compiles the less code to CSS and that's just what is exported. (~300kb saved right there!)
- Finally, enables stricter typings on the app, minimizing the surface area for new bugs to appear


libraries/ts/bonds/wasm-utils/Cargo.toml
```diff
@@ -3,6 +3,10 @@ name = "wasm-utils"
 version = "0.1.0"
 edition = "2021"
 
+# wasm-pack currently fails to install the correct binaryen version, so we disable wasm-opt for the time being
```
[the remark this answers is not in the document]
davodesign84: On the current bonds library is ~150kb vs ~110kb :)

app/src/utils/currency.ts
```diff
@@ -48,7 +48,7 @@ export function useCurrencyFormatting() {
     total: number,
     fiatValues?: boolean,
     price?: number,
-    decimals?: number,
+    decimals: number = 6,
```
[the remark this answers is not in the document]
davodesign84: Good one, I'll do that

suscd (dismissed):

.github/workflows/check.yml
```diff
@@ -167,6 +137,9 @@ jobs:
             !target/idl
             !target/deploy
 
+      - name: Install wasm-pack
+        run: cargo install wasm-pack
```
dnut, line 141: Yet another system dependency to maintain? Personally I prefer the repo to manage as many of its dependencies as possible automatically. I understand that you've seen issues with this, where you had to manually install it, but in most cases that wasn't necessary (for example in this workflow). So doesn't this change just amplify the problem by making it necessary 100% of the time? Now every single build environment has to manually install it, and persistent environments have to maintain it as a system-wide package and ensure they're using the correct version.

Also I think if we're going to have system-wide dependencies, I would put them in the docker image, not install them during the github workflow. That should make this workflow faster, plus that way the builds are reproducible everywhere, not just in github.

.github/workflows/check.yml
```diff
@@ -7,36 +7,6 @@ on:
   pull_request:
 
 jobs:
-  anchor-test:
```
dnut, line 10: In general, I prefer maximizing test coverage. I don't really see much advantage of removing these tests. It doesn't cost us anything to spin up another github runner to run in parallel during the other tests. But if you really think these tests are totally redundant, or just testing unused code, then I'll trust your judgement about removing them.

.github/workflows/check.yml
davodesign84: That's an interesting point. Yes it would be better place in the docker image, however we would still need to manually install them in other environments (e.g. cloudflare). It's also a bit more cumbersome to update the docker image, this would get it through quicker. I'd suggest to push it as is and update the docker image  / remove from workflow as a later PR, if that's ok with you?

.github/workflows/check.yml
davodesign84: Was mentioned here https://jetprotocol.slack.com/archives/C03JA4J5U8N/p1669904612378389 and discussed through in the wasm sdk chat. The reasoning of removing them is two-fold A) they are already covered by the E2E tests and B) they will soon be redundant once we move to the wasm sdk. Does that address your concerns?

dnut (approved):

PR #488, opened 2022-12-01
Renaming program variables

qiqicodes:
rename the following:
- `fixed_term_market` program and libraries
- `duration` to `tenor`
- `market ticket` to `ticket`
- `obligation` to `term loan`
- `market manager` to `market`


dnut:
"jet-market" is the crate name but "fixed_market" is the name elsewhere. I would probably be more consistent. Ideally the names would be identical with the exception of the "jet" prefix and maybe swapping dashes with underscores. For example, the program could be referred to as "fixed_market", "jet_fixed_market", or "jet-fixed-market", depending on context. Even that much variation makes me a little uncomfortable, but I can see its purpose.

Also, I'm confused by the name "fixed market". The grammatical meaning implies that there are markets which are fixed. But what does it mean for a market to be fixed? The actual thing being fixed is not the market. It's the term and rate of the financial instrument that is issued by the program. I would only put an adjective before a noun if the adjective is intended to modify that noun.

I would probably use one of these:
1. fixed term fixed rate lending market(s)
2. fixed term lending market(s)
3. fixed rate lending market(s)
4. fixed term market(s)
5. fixed rate market(s)
6. lending market(s)

I prefer verbosity over ambiguity. But at a certain point you get diminishing returns. So I would probably pick 2 or 3, but any of these seem OK. "jet_market" is too ambiguous, in my opinion. You could argue that some of our existing programs are also markets, and I wouldn't be surprised if we add more programs containing markets.

programs/fixed_market/src/control/state.rs
```diff
@@ -23,25 +23,25 @@ pub struct BondManager {
...
-    /// The token mint for the underlying asset of the bond tickets
+    /// The token mint for the underlying asset of the market tickets
     pub underlying_token_mint: Pubkey,
     /// Token account storing the underlying asset accounted for by this ticket program
     pub underlying_token_vault: Pubkey,
-    /// The token mint for the bond tickets
-    pub bond_ticket_mint: Pubkey,
-    /// Mint owned by bonds to issue claims against a user.
+    /// The token mint for the market tickets
+    pub market_ticket_mint: Pubkey,
```
dnut, line 31: I wouldn't use the name "market_ticket" because the market and the tickets are independent concepts. The tickets happen to only be created in a market, but they only represent the concept of lending and could in theory exist independently of any market. I would probably call them just "tickets". Maybe something like "fixed_term_tickets" or "lender_tickets" could work too. Also the mint is for creating free tickets, as opposed to claim and split tickets, so the mint could also be named "free_ticket_mint", though that name may be a little confusing without understanding the definition of "free"

programs/fixed_market/src/control/state.rs
```diff
@@ -4,13 +4,13 @@ use serde::{ser::SerializeStruct, Deserialize, Serialize, Serializer};
...
-/// The `BondManager` contains all the information necessary to run the bond market
+/// The `MarketManager` contains all the information necessary to run the fixed market
 ///
 /// Utilized by program instructions to verify given transaction accounts are correct. Contains data
-/// about the bond market including the tenor and ticket<->token conversion rate
+/// about the fixed market including the tenor and ticket<->token conversion rate
 #[cfg_attr(any(feature = "cli", test), derive(Deserialize))]
 #[account(zero_copy)]
-pub struct BondManager {
+pub struct MarketManager {
```
dnut, line 13: The word "manager" seems kinda redundant now. I would probably just call this "Market"

programs/fixed_market/src/control/state.rs
qiqicodes: updating `market ticket` to `ticket`

davodesign84 (approved):
I pushed a couple of tweaks to fix the build and looks good on the FE side. Perhaps @suscd or @dnut  want to have another look and push their approval tho.

davodesign84:
> @qiqicodes @davodesign84 I see that a rebase is needed. We shouldn't merge without rebasing or merging master so that we don't break builds immediately when merging.

I just pushed the button, let me know if that's ok for you :D

tests/scripts/localnet_lib.sh
```diff
@@ -81,7 +81,7 @@ start-oracle() {
 }
 
 start-crank-service() {
-    cargo run --bin jet-markets-crank-service -- --config-path $PWD/localnet.toml --verbose ${CRANK_VERBOSITY:0} &
+    cargo run --bin jet-markets-crank-service -- --config-path $PWD/localnet.toml --verbose ${CRANK_VERBOSITY:=0} &
```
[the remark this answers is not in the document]
davodesign84: This is not something I'm 100% confident on, but as far as I remember the `:=` operator assigns the value if itself if unset, the `:-` operator substitutes it if unset. Though I imagine that in some circumstances locally the `:-` operator can't assign it if it hasn't been declared first. The only issue with `:=` is that can't be used in certain cases, like positional parameters, but this is not the case. The other option might be to just declare the env variable at. the top of the file? This seemed more succint.

zachanon (dismissed):
Really nice @qiqicodes

programs/fixed_term/src/lib.rs
```diff
@@ -96,28 +95,28 @@
 //! the slot it was minted. To create a `SplitTicket`, you must configure your [`OrderParams`](struct@crate::orderbook::state::OrderParams) `auto_stake` flag to
 //! `true`. This will allow to program to immediately stake your tickets as the match event is processed.
 //!
-//! After the bond market tenor has passed, the ticket may be redeemed for the underlying value with the program. Also included are instructions
+//! After the Jet market market tenor has passed, the ticket may be redeemed for the underlying value with the program. Also included are instructions
```
dnut, line 98:
```suggestion
//! After the tenor has passed, the ticket may be redeemed for the underlying value with the program. Also included are instructions
```

programs/fixed_term/src/lib.rs
```diff
@@ -96,28 +95,28 @@
...
 //!
-//! When using a `jet-margin` account to post a collateralized borrow order, an [`Obligation`](struct@crate::orderbook::state::debt::Obligation) is created to track
-//! amounts owed to the program. `Obligation`s are either repaid manually by the user, or handled by an off-chain liquidator.
+//! When using a `jet-margin` account to post a collateralized borrow order, an [`TermLoan`](struct@crate::orderbook::state::debt::TermLoan) is created to track
+//! amounts owed to the program. `TermLoan`s are either repaid manually by the user, or handled by an off-chain liquidator.
 
 /// Program instructions and structs related to authoritative control of the program state
 pub mod control;
-/// Program instructions, methods and structs related to the use of margin accounts with the bonds program
+/// Program instructions, methods and structs related to the use of margin accounts with the Jet market program
```
dnut, line 108:
```suggestion
/// Program instructions, methods and structs related to the use of margin accounts with the fixed term program
```

Anchor.toml
```diff
@@ -13,7 +13,7 @@ skip-lint = true
 
 [programs.devnet]
 jet_test_service = "JPTSApMSqCHBww7vDhpaSmzipTV3qPg6vxub4qneKoy"
-jet_bonds = "JBond79m9K6HqYwngCjiJHb311GTXggo46kGcT2GijUc"
+jet_market = "JBond79m9K6HqYwngCjiJHb311GTXggo46kGcT2GijUc"
```
dnut, line 16:
```suggestion
jet_fixed_term = "JBond79m9K6HqYwngCjiJHb311GTXggo46kGcT2GijUc"
```

Anchor.toml
```diff
@@ -23,7 +23,7 @@ jet_metadata = "JPMetawzxw7WyH3qHUVScYHWFBGhjwqDnM2R9qVbRLp"
 
 [programs.localnet]
 jet_test_service = "JPTSApMSqCHBww7vDhpaSmzipTV3qPg6vxub4qneKoy"
-jet_bonds = "JBond79m9K6HqYwngCjiJHb311GTXggo46kGcT2GijUc"
+jet_market = "JBond79m9K6HqYwngCjiJHb311GTXggo46kGcT2GijUc"
```
dnut, line 26:
```suggestion
jet_fixed_term = "JBond79m9K6HqYwngCjiJHb311GTXggo46kGcT2GijUc"
```

dnut (dismissed):
my main concerns are:
- The crate "jet-market" and the folder "fixed_term" are totally different names which I see no reason for. I think this is only going to cause confusion. They should have some consistency like folder name = "some_program" and crate name = "jet-some-program"
- "market" is not a descriptive enough name for the fixed term program since we have other programs that are also markets, and the distinguishing characteristic of this program has nothing to do with markets but rather the fixed rate/fixed term style of lending and borrowing

suscd (approved):

PR #490, opened 2022-12-02
Remove unused margin swap code from library

nevi-me:
We don't need all this code as it interacts with the SPL program and not the margin program.


davodesign84 (changes requested):
I think you have forgotten to remove it from the top level index.ts :)

davodesign84 (approved):

PR #481, opened 2022-11-26
Update for upstream pyth changes

nevi-me:
`get_current_price()` is now `get_price_no_older_than()`, same with EMA.

PR #492, opened 2022-12-03
Route swap instruction

nevi-me:
This isolates the Rust changes for route swap instructions, so they can be reviewed separately.

#481 has the pyth changes, and can be reviewed separately to reduce the surface of this PR.

There are unfortunately dependencies that we'd have to point to git for some time until new versions are released or potential crate clashes are resolved.


programs/margin-swap/src/instructions/route_swap.rs
```diff
@@ -0,0 +1,528 @@
...
+        // CHECK: The swap program validates the below accounts
+        let swap_pool = &accounts[0];
+        let authority = &accounts[1];
+        let vault_into = &accounts[2];
+        let vault_from = &accounts[3];
+        let token_mint = &accounts[4];
+        let fee_account = &accounts[5];
+
+        // CHECK: The swap program gets validated by use_client! below
+        let swap_program = &accounts[6];
```
suscd, line 144: might be better to use `SwapInfo::try_accounts` to parse the accounts into identifiers

programs/margin-swap/src/instructions/route_swap.rs
```diff
@@ -0,0 +1,528 @@
...
+            ),
+            change_kind,
+            amount,
+        )?;
+
+        Ok(())
+    }
+
+    #[inline(never)]
+    fn spl_swap(
```
suscd, line 126: might be better to define the specific swap implementations in the instruction handler file for that swap type (e.g. this would be defined in `margin_spl_swap.rs`)

programs/margin-swap/src/instructions/route_swap.rs
```diff
@@ -0,0 +1,578 @@
...
+                    ctx.accounts
+                        .source_margin_pool
+                        .margin_pool
+                        .to_account_info(),
+                    ctx.accounts.source_margin_pool.vault.to_account_info(),
+                    ctx.accounts
+                        .source_margin_pool
+                        .deposit_note_mint
+                        .to_account_info(),
+                ];
```
suscd, line 223: Seems like it would be better to handle the withdrawal/deposits for tokens with the margin pool as a pre/post operation, instead of along with every swap leg:

* Client implementation is simplified somewhat, as constructing the route list no longer requires knowledge about if the swap is with a pool or not.
* Gives the client opportunity to allow users to have intermediate tokens that may not have a lending pool, by allowing the client to control the list of destination accounts that are deposited into the pool.

suscd (approved):

PR #493, opened 2022-12-07
 Turborepo example

davodesign84:


davodesign84:
We need a package there in order to execute yarn, otherwise it fails saying that there is no such package when on a fresh install. We could in theory have a palceholder package.json committed. Perhaps. But I don't see any harm having this committed for the time being, as long as it gets built.

Sent from Yahoo Mail on Android 
 
  On Thu, 8 Dec 2022 at 20:19, ***@***.***> wrote:   


@nevi-me commented on this pull request.

In packages/wasm/jet_lab_wasm_src_bg.wasm.d.ts:
> @@ -0,0 +1,73 @@
+/* tslint:disable */
+/* eslint-disable */
+export const memory: WebAssembly.Memory;

Is this something we should be commiting or should we be generating this as part of the build process?
The same with the wasm file

davodesign84:
Waiting on @https://github.com/jet-lab/jet-v2/pull/488 to get merged before updating this one

PR #500, opened 2022-12-12
Add more logging to the crank service

zachanon:
Adds dedicated logging logic so that maintenance of the crank-service is more manageable.


zachanon:
stale as of https://github.com/jet-lab/jet-v2/pull/505

PR #502, opened 2022-12-12
Restrict permission to refresh position metadata

dnut:
Only authorized users are allowed to refresh position metadata, instead of anyone. Previously there was no signer requirement. This repurposes the logic for liquidator approvals to a generic permit concept inside of margin that is used for both types of authorizations.

worth a second look: The airspace validation is a little weird since this is building off the liquidator authorization, which was already airspace scoped. That scope was not actually enforced because margin accounts have no concept of airspaces. So a liquidator or refresher may be authorized for a specific airspace, but they are actually allowed to act on any margin account. This needs margin accounts to be airspace-scoped before the authorization can actually limit refreshers to a particular airspace. Some alternative approaches are:
- using Pubkey::default() as a generic "approved everywhere" concept but I tried this and it increases complexity significantly in the program and especially in the libraries.
- instead of allowing them to act in every airspace, we could block them from working anywhere, until margin accounts are airspace aware


programs/margin/src/state/config.rs
```diff
@@ -138,19 +139,63 @@ pub enum TokenAdmin {
...
+}
+
+impl Permit {
+    pub fn validate(
+        &self,
+        airspace: Pubkey,
+        owner: Pubkey,
+        permissions: Permissions,
+    ) -> Result<()> {
+        // todo pubkey default is not acceptable once airspaces are in use
```
suscd, line 163: mark with `// FIXME: `

programs/margin/src/state/config.rs
```diff
@@ -164,6 +209,10 @@ pub struct AdapterConfig {
     pub adapter_program: Pubkey,
 }
 
-impl AdapterConfig {
-    pub const SPACE: usize = 8 + std::mem::size_of::<Self>();
+pub trait StorageSpace {
+    const SPACE: usize;
+}
```
suscd, line 214: seems generally useful for any program, maybe put this in the `program-common` library instead

programs/margin/src/state/config.rs
```diff
@@ -138,19 +139,63 @@ pub enum TokenAdmin {
...
+    /// airspace authority before an address is authorized to sign for the
+    /// instruction performing this action.
+    #[derive(Default, AnchorSerialize, AnchorDeserialize)]
+    #[repr(transparent)]
+    pub struct Permissions: u32 {
+        /// Liquidate margin accounts in this airspace.
+        const LIQUIDATE                 = 1 << 0;
+
+        /// Execute update_position_metadata for margin accounts in this airspace.
+        const REFRESH_POSITION_METADATA = 1 << 1;
```
suscd, line 197: should instead be named `REFRESH_POSITION_CONFIG`, since the metadata program will be phased out

programs/margin/src/lib.rs
```diff
@@ -500,11 +500,15 @@ pub mod jet_margin {
...
     /// The account storing the configuration will be funded if not already. If a `None` is provided as
     /// the updated configuration, then the account will be defunded.
-    pub fn configure_liquidator(
-        ctx: Context<ConfigureLiquidator>,
-        is_liquidator: bool,
+    pub fn configure_liquidator(ctx: Context<ConfigurePermit>, is_liquidator: bool) -> Result<()> {
+        configure_permit(ctx, is_liquidator, Permissions::LIQUIDATE)
+    }
+
+    pub fn configure_position_metadata_refresher(
```
suscd, line 507: name should be `configure_position_config_refresher` instead

programs/margin/src/state/config.rs
dnut: what difference does it make? looks like we have a lot todos and fixmes in the repo now

libraries/rust/margin/src/ix_builder/margin.rs
```diff
@@ -240,13 +246,19 @@ impl MarginIxBuilder {
...
+        position_token_mint: &Pubkey,
+        refresher: Pubkey,
+    ) -> Instruction {
+        let config = MarginConfigIxBuilder::new(self.airspace, self.payer, None)
             .derive_token_config(position_token_mint);
 
         let accounts = ix_account::RefreshPositionConfig {
             config,
             margin_account: self.address,
+            permit: derive_margin_permit(&self.airspace, &refresher),
```
suscd, line 260: this should just assume the payer has the permit

tools/ctl/src/actions/margin.rs
```diff
@@ -155,7 +155,7 @@ pub async fn process_refresh_metadata(client: &Client, token: Pubkey) -> Result<
                 plan = plan.instructions(
                     [],
                     [format!("refresh-position-md {token} for {address}")],
-                    [ix.refresh_position_metadata(&deposit_token)],
+                    [ix.refresh_position_metadata(&deposit_token, account.owner)],
```
suscd, line 158: using the account owner here is wrong, it should be the actual user signing, the account with the permit is probably always going to also be the payer

libraries/rust/margin/src/ix_builder/margin.rs
dnut: This is a bit inconvenient for testing. I'm trying to avoid using the payer for anything other than a faucet. In tests, I don't want it to be an authority of anything other than its own lamports. I'm trying to use separate keypairs for any signer roles. This makes it easier to track addresses while debugging, and enables stronger negative testing of permissions.

I see your point though, it's annoying to have to pass in the refresher here. It's already clear that the refresher is the actor, so it's reasonable to expect this data is included in the builder. I'm gonna mess around with some constructors to see if I can get something intuitive and easy to use that doesn't require a refresher here. One approach would be to just fund the refresher account I'm making in tests, and give it its own builder. Another would be to repurpose the signer here or to add an actor field.

dnut: the same struct also has this analogous method, where the liquidator needs to be passed in.
```rust
pub fn liquidator_invoke(&self, adapter_ix: Instruction, liquidator: &Pubkey) -> Instruction {
```

suscd: Yes the `liquidator_invoke` is a similarly awkward API

Seems like funding the refresher account is a simple and intuitive option. 

Even from the testing perspective, it just seems weird because the way its written can imply that its each user that is driving the refresh for their positions, whereas its really now a separate entity performing the action.

suscd (approved):

PR #503, opened 2022-12-13
Full flow settle

davodesign84:
This PR replaces the stats placeholder from the market selector with a conditional component that will allow the user to settle funds from filled borrowed orders.

In a following PR this is also the location where we will display a repay button for fixed market. I'm thinking that due to the lack of space, we can show the settle button first and if nothing is left to settle, we can show the repay button for now. Until we have a better solution for the UI.


eyevz (approved):
This is a reasonable approach, and is working as advertised.

PR #505, opened 2022-12-13
fixed-term: refactor derivation for loan/deposit accounts

suscd:
- merge `SplitTicket` and `ClaimTicket` into `TermDeposit`
- make derivation for loan/deposit tracking accounts deterministic, depending on a sequence number tracked in the `MarginUser` account
- rewrite client-side event consumer to account for need to track user accounts to determine the correct sequence numbers
- fix memory issues in `consume_events`


eyevz:
Haven't been through all the code yet, but extensive manual testing of the UI was very positive - the app feels pretty robust.

programs/fixed-term/src/margin/state.rs
```diff
@@ -39,22 +39,22 @@ pub struct MarginUser {
...
     /// Amount that must be collateralized because there is an open order for it.
     /// Does not accrue interest because the loan has not been received yet.
-    pending: u64,
+    pub pending: u64,
 
     /// Debt that has already been borrowed because the order was matched.
     /// This debt will be due when the loan term ends.
     /// This includes all debt, including past due debt
-    committed: u64,
+    pub committed: u64,
```
dnut, line 57: Do this all really need to be public? The idea of making it private was to encapsulate the complexity of managing this state. It's nice to get a compiler guarantee that it's not being touched anywhere in the program so you can reason about the behavior of this struct purely by looking at its methods. Setting aside `next_new_term_loan_seqno` for a moment, none of the other fields are used at all anywhere else. Why shouldn't they be private?

programs/fixed-term/src/orderbook/instructions/consume_events/queue_iterator.rs
```diff
@@ -9,33 +9,33 @@ use crate::{
...
 pub struct EventIterator<'a, 'info> {
     queue: QueueIterator<'info>,
     accounts: Iter<'a, AccountInfo<'info>>,
-    /// CHECK: anchor linter bug requires this
     system_program: AccountInfo<'info>,
-    /// CHECK: anchor linter bug requires this
     payer: AccountInfo<'info>,
-    seeds: std::vec::IntoIter<Vec<u8>>,
+    market: Pubkey,
+    seed: Vec<u8>,
```
dnut, line 38: Imagine a single non-margin participant has two orders resting on the same order book. A borrower places an order as a taker and matches against both of those orders. This results in two consecutive events in the event queue that each require a term deposit to be created for the same fill account. If the crank attempts to consume both events in the same instruction, the execution will fail due to an address collision from using the same seeds.

programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// The tracking account for the deposit
+    #[account(mut,
+              close = owner,
+              has_one = owner
+    )]
+    pub deposit: Account<'info, TermDeposit>,
+
+    /// The account that owns the deposit
+    #[account(mut)]
+    pub owner: Signer<'info>,
```
dnut, line 21: A TermDeposit can be owned by a MarginUser which is not a signer. You've removed the distinction between "authority" and "ticket_holder", but that was actually necessary because they may be different. The previous approach had a parameter in the redeem function to specify the owner, and the same approach could work here.

programs/fixed-term/src/margin/instructions/margin_redeem_deposit.rs
```diff
@@ -3,13 +3,13 @@ use jet_program_proc_macros::MarketTokenManager;
...
-    tickets::instructions::redeem_ticket::*, FixedTermErrorCode,
+    tickets::instructions::redeem_deposit::*, FixedTermErrorCode,
 };
 
 #[derive(Accounts, MarketTokenManager)]
-pub struct MarginRedeemTicket<'info> {
+pub struct MarginRedeemDeposit<'info> {
     #[account(mut,
-		constraint = margin_user.margin_account == inner.authority.key() @ FixedTermErrorCode::WrongMarginUserAuthority,
+		constraint = margin_user.margin_account == inner.owner.key() @ FixedTermErrorCode::WrongMarginUserAuthority,
```
dnut, line 12:
```suggestion
		address = inner.owner.key() @ FixedTermErrorCode::WrongMarginUserAuthority,
```

programs/fixed-term/src/margin/state.rs
```diff
@@ -168,6 +168,12 @@ pub struct Assets {
     /// tickets to transfer into settlement account
     pub entitled_tickets: u64,
 
+    /// The sequence number for the next deposit
+    pub next_deposit_seqno: u64,
+
+    /// The sequence number for the oldest deposit that has yet to be redeemed
+    pub next_unredeemed_deposit_seqno: u64,
```
dnut, line 175: The sequence is not being enforced for redemption, so this number is not accurate. It's also not necessary to enforce it so it should be fine to just remove this, unless you'd like to actually enforce it to provide some guarantees on the client-side.

programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
```diff
@@ -0,0 +1,87 @@
...
+                self.token_program.to_account_info(),
+                Transfer {
+                    to: self.token_account.to_account_info(),
+                    from: self.underlying_token_vault.to_account_info(),
+                    authority: self.market.to_account_info(),
+                },
+            )
+            .with_signer(&[&self.market.load()?.authority_seeds()]),
+            self.deposit.amount,
+        )?;
```
dnut, line 69: if you derive MarketTokenManager this could actually be:
```suggestion
        self.withdraw(
            self.underlying_token_vault.to_account_info(),
            self.token_account.to_account_info(),
            self.deposit.amount,
        )?;
```

programs/fixed-term/src/tickets/instructions/transfer_deposit.rs
```diff
@@ -0,0 +1,25 @@
+use anchor_lang::prelude::*;
+
+use crate::tickets::{events::DepositTransferred, state::TermDeposit};
+
+#[derive(Accounts)]
+pub struct TransferDeposit<'info> {
+    /// The deposit to transfer
+    #[account(mut, has_one = owner)]
```
dnut, line 8: This has the same problem as redemption where the margin user cannot sign, so they won't be able to transfer term deposits. Looks like they had the same problem with the prior implementation of this instruction though, so this isn't actually a regression here.

tests/scripts/localnet_lib.sh
```diff
@@ -81,7 +81,11 @@ start-oracle() {
 }
 
 start-crank-service() {
-    cargo run --bin jet-fixed-terms-crank-service -- --config-path $PWD/localnet.toml --verbose ${CRANK_VERBOSITY:=0} &
+    log_filter="jet_margin_sdk=debug"
+    env RUST_LOG=$log_filter cargo run --bin jet-fixed-terms-crank-service -- \
+        --config-path localnet.toml \
+        --log-path .localnet/crank.log \
```
dnut, line 87: this results in a file not found error if `.localnet/` does not exist

programs/fixed-term/src/orderbook/instructions/consume_events/queue_iterator.rs
suscd: Changed this to increment the seed to avoid using the exact same seed for each event.

programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
suscd: Fixed to include an `authority` account, which allows the owner to be different from the required signer.

programs/fixed-term/src/margin/state.rs
suscd: this is now enforced, so that clients can rely on it

programs/fixed-term/src/tickets/instructions/transfer_deposit.rs
suscd: decided to do nothing here, and accept that margin account's won't be able to transfer their deposits

dnut (approved):

PR #514, opened 2022-12-16
Change events to match `OrderTag` instead of `order_id`

zachanon:
The `OrderTag` type is unique and deterministic, and so should be what we use to track specific order events as they move around our programs.


davodesign84 (approved):
Looks good to me

dnut (approved):

PR #515, opened 2022-12-16
Add order information needed for a crank to "auto roll" orders

zachanon:
We want users to be able to specify that a crank should be allowed to "roll" their expiring `TermDeposit` or `TermLoan` by placing a new order to the market.

This PR is adding the necessary data to orders to allow this instruction to function


programs/fixed-term/src/orderbook/state.rs
```diff
@@ -101,14 +101,19 @@ impl<'info> OrderbookMut<'info> {
         flags: CallbackFlags,
     ) -> Result<(CallbackInfo, SensibleOrderSummary)> {
         let mut manager = self.market.load_mut()?;
+        let auto_roll = if params.auto_roll {
```
dnut, line 104: This function is used by five instructions but I think auto-roll only makes sense for margin borrow and lend at this point.

programs/fixed-term/src/margin/instructions/configure_auto_roll.rs
```diff
@@ -0,0 +1,56 @@
...
+        mut,
+        has_one = margin_account,
+    )]
+    pub margin_user: Box<Account<'info, MarginUser>>,
+
+    /// The signing authority for this user account
+    #[account(
+        constraint = margin_account.owner == &jet_margin::ID,
+    )]
+    pub margin_account: Signer<'info>,
```
suscd, line 32: this is the wrong way to validate a margin account

```suggestion
    #[account(signer)]
    pub margin_account: AccountLoader<'info, MarginAccount>,
```

suscd (approved):

PR #518, opened 2022-12-16
consolidate libs, ensure markets are refreshed during pool operations…

davodesign84:
… and add e2e test to cover for potentially stale market positions.
Also adds the fixed-term idl to the prettier ignore as per @dnut request


packages/margin/src/fixed-term/fixedTerm.ts
```diff
@@ -392,30 +392,30 @@ export class FixedTermMarket {
   }
 
   async deriveMarginUserAddress(user: MarginAccount): Promise<PublicKey> {
-    return await findDerivedAccount(["margin_borrower", this.address, user.address], this.program.programId)
+    return await findFixedTermDerivedAccount(["margin_borrower", this.address, user.address], this.program.programId)
```
zachanon, line 395: If we provide the fixed term program id, I am not sure on the clarity of adding `findFixedTermDerivedAccount`, it would be the same method regardless of program. Just a nitpick.

packages/margin/src/margin/marginAccount.ts
```diff
@@ -1745,4 +1747,55 @@ export class MarginAccount {
...
+       const borrowPosition = this.getPositionNullable(pool.addresses.loanNoteMint)
+       if (pool.address === this.address) {
+         poolsToRefresh.push(pool)
+       } else if (borrowPosition && !borrowPosition.balance.eqn(0)) {
+         poolsToRefresh.push(pool)
+       } else if (assetCollateralWeight > 0 && depositPosition && !depositPosition.balance.eqn(0)) {
+         poolsToRefresh.push(pool)
+       }
+     }
+     console.log(instructions.length)
```
zachanon, line 1793: Curious why we log this to the console

davodesign84: Ah good spot I'll remove it

PR #525, opened 2022-12-21
Generate events on order canceling

zachanon:
Events were not being generated on order cancels, which meant user funds were irretrievable.


eyevz (approved):
Manual testing show this working as expected so let's merge it for now.

See @dnut's ticket JET-1162 and also @suscd's comment that it may be preferable not to use the event queue if that isn't necessary.

suscd (approved):

PR #529, opened 2022-12-23
Allow jet-lab-wasm-src to depend on jet-fixed-term

eyevz:
It would be neater for jet-lab-wasm-src to depend on the agnostic orderbook and jet fixed term program, instead of copying in code.

A dependency on the aaob was simple.

Depending on jet-fixed-term was more complicated because of its dependency on the margin program whose AccountPositionKey and AccountPositionList structs have usize fields. The wasm target is a 32-bit format and I unable to resolve rust error E0512, presumably from bytemuck, when building for wasm.

So I replaced the usize fields with u64 fields and made the downstream changes.


libraries/rust/program-common/src/index_value.rs
```diff
@@ -0,0 +1,57 @@
+use std::ops::{AddAssign, SubAssign};
+
+use anchor_lang::{AnchorDeserialize, AnchorSerialize};
+use bytemuck::{Pod, Zeroable};
+
+#[derive(Debug, Default, AnchorDeserialize, AnchorSerialize, Clone, Copy, Pod, Zeroable)]
+#[repr(C)]
+pub struct IndexValue(pub u64);
```
suscd, line 8: I'm not sure what the point of this type is? Why not just change `usize` to `u64`?

eyevz: The operation for going from `u64` to `usize` for indexing operations depends on the target architecture, and I wanted to encapsulate that somehow. Since I can't directly `impl From<u64> for usize`, I used `IndexValue` to wrap the `u64`. This allows me to condition the implementation of `from()` on the target architecture.

eyevz: I guess I could change to `u64` directly, and simply have cfg-conditional implementations for a function like `u64_to_usize()`. Preferable?

suscd: usize already implements `TryFrom<u64>`, so you can use `usize::try_from` to force an error if the value is actually too big to use as an index

eyevz: Yeah much simpler that way.

eyevz:
Let's overlook the codecov for now - it's from a work-in-progress wasm-src file.

suscd (approved):

PR #538, opened 2023-01-05
Settle users with crank

dnut:
Users don't have to execute "settle" any more. It's handled by the crank that processes the orderbook event queue.

# Background

## `consume_events` limitation
The consume_events instruction processes each orderbook event in sequence. The way it is currently implemented requires each event to be processed in full before any future events are processed. It is completely serial and cannot benefit from any sort of parallelization. It breaks if there is any out-of-order execution. There can only be a single CPU in the world at a time that is running this crank. If the consume_events instruction suffers from errors or slowdowns then it grinds the orderbook to a halt. So ideally we can make this instruction as lean and simple as possible to avoid any issues, and push other work to processes that can be executed at any time, in any order, and can benefit from parallelization.

## `settle` instruction purpose
The settle instruction basically only serves to clean up state that was left behind by consume_events. All of this logic could easily be put into consume_events, which is handled by a crank, and then no one would need to settle anything.

The problem is that it makes the consume_events instruction more complex and it requires a lot more accounts, which is a burden on the single serialized crank that has to manage it. So the settle instruction was created to extract out some of the work that can be done at any time, in any order, and can benefit from serialization. The crank can run through events as fast as possible but it leaves user accounts in an incomplete state that can be cleaned up whenever.

Currently we've made it the user's responsibility to execute settle, but there is no particular need for this. It could also be handled by the crank, or some other crank, or some collection of other cranks running in parallel.

# This change
The crank that processes events can easily figure out when accounts need to be settled. If there was an event for an account, then it needs to be settled, simple as that. This change has the crank keep track of those accounts and send out settle transactions in a relaxed manner without blocking the processing of the event queue.

## Implementation
This adds a deduplicated queue within the crank where it pushes accounts that need to be settled. There is another thread that occasionally pops a group of accounts off the queue and sends out a settle transaction for each one. It sends them all at once without blocking anything waiting for responses.


davodesign84:
I'm wondering, how would the UI know when and how much the user should withdraw to the pool?
If I understand correctly once the settle instruction is called, the entitledTokens key in the user assets is reset to 0 and the funds are moved to the margin account, of which the user has currently no visibility on.

Is the intent to merge this as part of some additional amount of work that includes the management of funds attached to the user margin account?

davodesign84:
I generally agree with the approach. I'm just a bit annoyed that the we would now need to rework the ui to take into account funds in the margin account (but we know that's needed anyway, eventually).

Just to clarify then, would this mean that the front-end will never have to be concerned about settling and instead would just look at outstanding funds in the pda?

dnut:
> Just to clarify then, would this mean that the front-end will never have to be concerned about settling and instead would just look at outstanding funds in the pda?

This is the general direction of the change, but settlement is not perfectly guaranteed by the crank.

If the settlement instruction does not complete successfully, the crank will keep retrying to settle that account forever. But if there's a bug in the crank that causes failure every time, then the account will remain stale.

If the crank is turned off while some accounts remain unsettled, it currently has no way to re-learn about those accounts. Currently it only decides to settle an account when it processes an event for that account. It does not scan other accounts looking for ones that need to be settled. We could implement something like that if we want stronger assurances for the frontend, but it doesn't exist right now.

Of course there's always the possibility that we rethink our usage of cranks at which point we'll be looking for alternatives.

programs/fixed-term/src/margin/state.rs
```diff
@@ -85,22 +85,22 @@ impl MarginUser {
 #[derive(Zeroable, Debug, Clone, AnchorSerialize, AnchorDeserialize)]
 pub struct Debt {
     /// The sequence number for the next term loan to be created
-    next_new_term_loan_seqno: u64,
+    pub next_new_term_loan_seqno: u64,
```
suscd, line 88: don't think these need to be made public, don't appear to be used anywhere else in this change?

dnut: weird, I don't remember doing this

suscd (approved):

PR #542, opened 2023-01-09
Significantly refactored approach to wasm helpers for fixed term

eyevz:
* Only the orderbook-related part of the lib has been split into independent Rust code
  and separate binding code. This works quite well (see below) and we should probably
  bring the rest of the wasm lib to this format.
* Typescript bindings are not generated because we're now returnign JSON from data-
  generating methods. This is work-in-progress. wasm-bindgen apears to have some
  mechanisms for overriding the Typescript definitions output.


packages/margin/src/fixed-term/fixedTerm.ts
```diff
@@ -112,6 +111,7 @@ export class FixedTermMarket {
   }
   readonly info: MarketInfo
   readonly program: Program<JetFixedTerm>
+  public orderbookModel: OrderbookModel | undefined = undefined
```
davodesign84, line 114: you could use the shorthand `public orderbookModel?: OrderbookModel`

packages/margin/src/fixed-term/fixedTerm.ts
```diff
@@ -457,8 +457,16 @@ export class FixedTermMarket {
     return await findFixedTermDerivedAccount(["term_deposit", this.address, marginUser, seed], this.program.programId)
   }
 
-  async fetchOrderbook(): Promise<Orderbook> {
-    return await Orderbook.load(this)
+  async fetchOrderbook(tenor: bigint): Promise<OrderbookModel> {
+    const asksBuf = (await this.provider.connection.getAccountInfo(this.info.asks))!.data
```
davodesign84, line 461: this will throw an exception if the asks can't be accessed, creates issue during e2e tests

davodesign84 (changes requested):
Couple of small things to address but looks good

davodesign84 (dismissed):

suscd (approved):

suscd (approved):

PR #544, opened 2023-01-09
misc program changes

suscd:
various minor code changes to accommodate the new client library:

- changed tenor types from `i64` to `u64`, since tenor cannot be negative
- remove ticket price update from fixed-term refresh
- add a `margin_borrow_v2` instruction to pools which allows tokens to be immediately sent to the user's account without a separate withdraw
- fix `transfer_deposit` for withdrawal case

the test for `margin_borrow_v2` is missing from this change, as its coupled with the new client code to be introduced in a another PR (#545)


eyevz (approved):
This looks okay to me.

dnut (approved):

PR #549, opened 2023-01-11
Redeem deposit

davodesign84:


davodesign84:
@dnut @suscd @zachanon I'm happy with the branch, could you have a review as your approval is required for merging.

dnut:
I made the program changes, so I'd like to get a second opinion on them. @suscd I'm curious what you think about making register_position return Ok if the position already exists. It just leaves the position alone and makes no changes. Previously it returned an error. This affects multiple instructions.

suscd (approved):

suscd:
lgtm

PR #550, opened 2023-01-11
settle fixes

dnut:
This adds tighter constraints to the fixed term program. Standard practices for settlement are now hard requirements. This is to avoid accounting issues in margin. Libraries are updated to conform to new expectations.

Always use accounting invoke - reduces risk to user
- Settle can now only be invoked through margin, enforced by return to margin
- settler crank now uses accounting_invoke, previously it did not, which may have caused some confusion to users of the frontend who are relying on the crank

Always settle to ATAs - reduces risk to system
- no longer configurable, the ata is enforced at time of settle 
- Settle ix requires ATA to be registered as a position if its going to be moving anything in there

misc
- alphabetize errors
- fix issue in margin requiring mut self when not needed
- register underlying and ticket as deposits during fixed term tests
- update idl

### DIFF
--- a/programs/fixed-term/src/margin/instructions/margin_borrow_order.rs
+++ b/programs/fixed-term/src/margin/instructions/margin_borrow_order.rs
@@ -1,6 +1,6 @@
 use agnostic_orderbook::state::Side;
 use anchor_lang::prelude::*;
-use anchor_spl::token::Token;
+use anchor_spl::{associated_token::get_associated_token_address, token::Token};
 use jet_margin::{AdapterResult, PositionChange};
 use jet_program_common::traits::{SafeSub, TryAddAssign};
 use jet_program_proc_macros::MarketTokenManager;
@@ -59,8 +59,11 @@ pub struct MarginBorrowOrder<'info> {
     #[account(mut, address = orderbook_mut.vault() @ FixedTermErrorCode::WrongVault)]
     pub underlying_token_vault: AccountInfo<'info>,
 
-    /// The market token vault
-    #[account(mut, address = margin_user.underlying_settlement @ FixedTermErrorCode::WrongUnderlyingSettlementAccount)]
+    /// Where to receive borrowed tokens
+    #[account(mut, address = get_associated_token_address(
+        &margin_user.margin_account,
+        &orderbook_mut.underlying_mint(),
+    ))]
     pub underlying_settlement: AccountInfo<'info>,
 
     #[market]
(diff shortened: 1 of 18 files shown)


programs/fixed-term/src/margin/instructions/settle.rs
```diff
@@ -53,11 +60,19 @@ pub struct Settle<'info> {
     /// CHECK: token program checks it
     #[account(mut)]
     pub ticket_mint: AccountInfo<'info>,
-    /// CHECK: token program checks it
-    #[account(mut)]
+
+    /// Where to receive owed tokens
+    #[account(mut, address = get_associated_token_address(
```
dnut, line 65: I tried using anchor's `associated_token` syntax here but it wouldn't compile. Seems like a bug in anchor.

programs/fixed-term/src/margin/instructions/margin_borrow_order.rs
```diff
@@ -59,8 +59,11 @@ pub struct MarginBorrowOrder<'info> {
     #[account(mut, address = orderbook_mut.vault() @ FixedTermErrorCode::WrongVault)]
     pub underlying_token_vault: AccountInfo<'info>,
 
-    /// The market token vault
-    #[account(mut, address = margin_user.underlying_settlement @ FixedTermErrorCode::WrongUnderlyingSettlementAccount)]
+    /// Where to receive borrowed tokens
+    #[account(mut, address = get_associated_token_address(
```
dnut, line 63: I tried using anchor's `associated_token` syntax here but it wouldn't compile. Seems like a bug in anchor.

dnut (commented):
some explanation

programs/fixed-term/src/margin/instructions/settle.rs
suscd: it only works with accounts as arguments, not pubkeys

programs/fixed-term/src/margin/instructions/margin_borrow_order.rs
suscd: FWIW I don't think you even need to check this here, because this instruction requires the margin account to sign, so the accounting should remain correct regardless of where the tokens are sent (unlike `settle` which is permissionless, yet alters the accounting).

Though I think its fine to keep the check for consistency, a client can always add a withdraw instruction if they need to (and we could always relax it later to make things like flash loan composability a little easier perhaps)

dnut: Yeah I was debating how to handle this instruction and just decided to follow the same pattern for now. I figure it's easier to loosen the restriction if we decide that would be valuable rather than tighten the restriction after some clients may already start using a different account for no reason.

dnut: One inconsistency with the general pattern though is that this instruction does not require that the position is registered. I guess it would be in the spirit of what I just said to actually require it. But it's not really needed for the same reason. And to me it doesn't seem as difficult to add that requirement in the future.

PR #552, opened 2023-01-12
fix `margin_lend_order` to set the right sequence numbers

suscd:


eyevz (approved):
Good spot, makes sense.

davodesign84 (approved):

PR #560, opened 2023-01-21
Make `MarginAccount` structs aware of airspaces

zachanon:


programs/margin/src/instructions/liquidate_end.rs
```diff
@@ -28,7 +28,9 @@ pub struct LiquidateEnd<'info> {
     pub authority: Signer<'info>,
 
     /// The account in need of liquidation
-    #[account(mut, has_one = liquidation)]
+    #[account(mut,constraint =
```
zachanon, line 31: Should be a 'has_one'

suscd: you should use `has_one` for the `liquidator` and `margin_account` fields for validation on the `liquidation` account

suscd (changes requested):
can you also add an airspace check on `register_position`? https://github.com/jet-lab/jet-v2/blob/mrgn/airspace-fields/programs/margin/src/state/account.rs#L177

since when using the `TokenConfig` accounts, they need to have the airspace match with the margin account

programs/margin/src/instructions/liquidate_end.rs
zachanon: I don't think we need the check on the liquidator if we confirm the margin account. In the instruction we verify the authority signer by checking the `liquidator` key in the margin account, so there is no way to slip in an unauthorized use.

programs/margin/src/instructions/liquidate_end.rs
```diff
@@ -28,11 +28,15 @@ pub struct LiquidateEnd<'info> {
...
+        margin_account.load()?.liquidator == liquidation.load()?.liquidator
+            @ ErrorCode::UnauthorizedLiquidator)]
     pub margin_account: AccountLoader<'info, MarginAccount>,
 
     /// Account to persist the state of the liquidation
-    #[account(mut, close = authority)]
```
suscd, line 35: looks like you accidentally removed the close here

programs/margin/src/instructions/register_position.rs
```diff
@@ -32,15 +30,15 @@ pub struct RegisterPosition<'info> {
     pub payer: Signer<'info>,
 
     /// The margin account to register position type with
-    #[account(mut)]
+    #[account(mut,
+        constraint = margin_account.load()?.airspace == config.airspace)]
```
suscd, line 34: you _really_ should be doing this check in the `register_position` function on the `MarginAccount` state, to ensure that every caller has to think about validating the airspace

libraries/rust/instructions/src/margin.rs
```diff
@@ -57,7 +61,7 @@ pub struct MarginIxBuilder {
 impl MarginIxBuilder {
     /// Create a new [MarginIxBuilder] which uses the margin account as the authority.
     /// Ordinary margin users should use this function to create a builder.
-    pub fn new(airspace: Pubkey, owner: Pubkey, seed: u16) -> Self {
+    pub fn new(airspace: String, owner: Pubkey, seed: u16) -> Self {
```
suscd, line 64: couldn't figure out the point of making everything use the airspace as a string (to have to derive the pubkey), since it doesn't seem like the string is actually used for anything at all?

suscd (changes requested):

libraries/rust/instructions/src/margin.rs
zacheryasc: The `AirspaceIxBuilder` requires the string, ran into a case where I didn't have the string just the address, decided to leave the instantiation of the builder alone and just make the string more accessible.

zachanon: The `AirspaceIxBuilder` requires the string. I ran into a case where I had access to the address and not the string, decided that the string was more information, since the address is derived from it, and so it was inherently more useful to pass around.

programs/margin/src/instructions/register_position.rs
suscd: I don't think any of that matters? 

You just have to get the airspace key in the `TokenConfig`

libraries/rust/instructions/src/margin.rs
suscd: Sure, but now you're just placing an extra burden on code that only ever care about the address (because they now have to  always take the extra step to derive it instead). That doesn't really sound like an improvement

zachanon: As a general code pattern, because I have run into similar issues of having the address but not the information used to derive it, I was thinking of something like

```
trait Addressable {
   fn address(&self) -> Pubkey;
}

struct AirspaceAddressBuilder: Addressable {
  name: String,
}

impl Addressable for AirspaceAddressBuilder { ... }
```

I did not build this pattern into the PR, but laid some foundation to implement it later. This is what I mean by "more information", because it may be useful to know the airspace name, the marginaccount seed, the fixed-term market seeds without fetching from the chain. I don't see this in resulting in significant complexity growth, in many cases we reduce it, as in this case where we needed an `AirspaceIxBuilder` and lacked the string, and we can slowly incorporate the pattern in the same way we have handled the `TransactionBuilder`.

zachanon: I see this as like having a filepath that you cannot walk back up

`/usr/lib/bin` vs `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`

very annoying to deal with constantly.

programs/margin/src/instructions/register_position.rs
zachanon: For example, when returning the data from a fixed term borrow
https://github.com/jet-lab/jet-v2/blob/master/programs/fixed-term/src/margin/instructions/margin_borrow_order.rs#L200

If we are adding an `airspace` field, then registering a position from the adapter means we need to include airspace information into the adapter call.

I can see:
- adapter call needs a `TokenConfig` for any registered position, use that as the `pubkey` in the `PositionChange` [enum](https://github.com/jet-lab/jet-v2/blob/master/programs/margin/src/adapter.rs#L83) (which then fetches the account info? or we send a `TokenConfig` instead of a pubkey)
- airspace is provided somehow alongside the pubkey for the `PositionChange`, so the enum variant looks like `PositionChange::Register(Pubkey, Pubkey) // airspace, token`
- adapter invoke requires some kind of airspace permit

suscd: I really think that you're overthinking this one, registering a position *always* requires reading the `TokenConfig` account. When registering from an adapter, the margin program still has to read the `TokenConfig`, which the adapter isn't required to know anything about. 

You only have to check that the airspace in the `TokenConfig` (which is already being read, and indeed *must* be read) matches the margin account

libraries/rust/instructions/src/margin.rs
suscd: I could maybe see that argument if the seed value was actually useful for anything, but it's not. The only time you actually need to know the seed is when creating an airspace, and at no point after. So at the end of the day, you're kind of doing all this for no benefit, because there is just no reason to care about the seed value for an airspace

PR #562, opened 2023-01-23
update dependencies

dnut:
**Problem:** Waiting until an update is necessary. Dependent work gets blocked. Update causes many build errors entangled in a mess of confounding factors.

**Solution:** Frequent trivial updates before anything is blocked. Issues are easier to resolve because they are isolated.

- updated github builder image with latest stable versions of all tools:
  - rust 1.66.1
  - node 18.13
  - solana 1.13.7
  - anchor 0.26
- fixed clippy errors that old clippy 1.65 fails to catch
- `cargo +bpf update`
- updated jetctl logic to be compatible with new idl account struct, manually tested with:
`cargo run --bin jetctl -- -um proposals inspect FUn4szLHM4AepgRru6ZgZBZ4RuosqKN66kRBFQcN5TsN`

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,7 +10,7 @@ jobs:
   hosted-tests-localnet-batch1:
     runs-on: ubuntu-latest
     container:
-      image: jetprotocol/builder:rust-1.65.0-node-18.12.1-solana-1.13.6-anchor-0.25.0-1
+      image: jetprotocol/builder:rust-1.66.1-node-18.13.0-solana-1.13.7-anchor-0.26.0-1
     env:
       HOME: /home/tools
       BATCH: batch1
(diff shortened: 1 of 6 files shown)


dnut:
> LGTM, only question's why not solana 1.14.x? Is it because it's not yet on mainnet?

Basically I just wanted this change to be a simple conservative update. Any opinionated changes to our approach to CI can go separately

The builder image has always used the stable release, which is still 1.13. This is basically only used for the test validator when tests are run. It is nice to be confident our code would work on mainnet before merging to master.

If we know we won't be deploying the code to mainnet soon, and there's a specific reason that we need to test against a later version, then we could change this, but we'll need to be careful about deploying that to mainnet. Maybe it would make sense to test the master and mainnet branches with different versions, but that's a more complicated setup.

PR #584, opened 2023-02-07
Collateralize Open Borrow Orders

zachanon:
To collateralize tokens locked into a borrow order:

- the program is now aware of `ticket_collateral` and `token_collateral`
- open orders are now counted as token assets


.github/e2e-docker-compose.yml
```diff
@@ -10,15 +10,15 @@ services:
     # ports:
     #   - 27017:27017
   postgres:
-    image: gcr.io/kubernetes-608504/jet-data-postgres:development
+    image: gcr.io/kubernetes-608504/jet-data-postgres:pr
```
dnut, line 13: Is this intented to be merged into master?

programs/fixed-term/src/control/state.rs
```diff
@@ -36,6 +36,10 @@ pub struct Market {
     /// positions that are priced as tickets. The collateral notes are monitored
     /// by the margin program to track value
     pub ticket_collateral_mint: Pubkey,
+    /// Mint owned by fixed-term market to issue collateral value to a user for
+    /// positions that are priced as tokens. The collateral notes are monitored
+    /// by the margin program to track value
+    pub token_collateral_mint: Pubkey,
```
dnut, line 42: In most places we're using the word "underlying". "Token" is ambiguous to me since there are a lot of different tokens here: underlying, claims, tickets, ticket_collateral, and this new collateral. My initial instinct for a name is `underlying_collateral_mint`

programs/fixed-term/src/margin/instructions/initialize_margin_user.rs
```diff
@@ -80,25 +91,22 @@ pub struct InitializeMarginUser<'info> {
...
-            claims: ctx.accounts.claims.key(),
-            ticket_collateral: ctx.accounts.ticket_collateral.key(),
-            borrow_roll_config: Default::default(),
-            lend_roll_config: Default::default(),
-        } ignoring {
-            debt,
-            assets,
-        }
-    }
+    ***user = MarginUser::new(
```
dnut, line 102: Curious why you got rid of the init macro. Its purpose was twofold:
1. Compile-time guarantee that we have explicitly specified how each field should be initialized.
2. Maximize performance by mutating the existing struct rather than replacing it with a new one.

I didn't see another way to get both of those benefits, without using some macro like this.

I considered using the constructor approach, like this here. It also ensures that every field is initialized explicitly. But it does so with a performance cost, though maybe that's negligible. It also has a risk that the function parameters change meaning without changing type, and the logic over here is not corrected. Named parameters minimize this risk. This may also be negligible, though it becomes more likely as functions have more parameters. I just didn't see any downsides to the init macro.

programs/fixed-term/src/margin/instructions/initialize_margin_user.rs
```diff
@@ -120,6 +128,12 @@ pub fn handler(ctx: Context<InitializeMarginUser>) -> Result<()> {
                         ctx.accounts.ticket_collateral.key(),
                     )],
                 ),
+                (
+                    ctx.accounts.token_collateral_mint.key(),
+                    vec![PositionChange::Register(
```
dnut, line 133: as we add more of these, the utility of an instruction to close a MarginUser increases

programs/fixed-term/src/margin/instructions/margin_sell_tickets_order.rs
```diff
@@ -48,10 +48,14 @@ pub fn handler(ctx: Context<MarginSellTicketsOrder>, params: OrderParams) -> Res
             .map(|a| a.key()),
         CallbackFlags::MARGIN,
     )?;
+
+    // collateral accounting
+    let posted_token_value = order_summary.quote_posted(RoundingAction::PostLend)?;
+    ctx.accounts.margin_user.sell_tickets(posted_token_value)?;
     ctx.mint(
```
dnut, line 55: This should remain ticket_collateral and quote quantity.

To represent their open order, this gives the user a collateral position equal in value to the underlying, but in the quantity of the tickets. Since underlying is always more valuable than tickets, this is actually more valuable than anything they could possibly get from this order. Let's say 110 tickets are worth about $100, and underling is worth about $1. If they're trying to sell 110 tickets for 100 underlying, they are trying to sell $100 worth of tickets for $100 worth of underlying, but they're granted $110 worth of collateral. It would be in a user's best interest to place a sell order of tickets for the highest price possible and they will be granted a massive collateral position that vastly exceeds the value of anything they have deposited. This can be used as an exploit to steal from lenders.

An open order is like a superposition of two states: fill + cancel + a wide range of prices. It can randomly resolve into any of these states. So, if we're going to treat the open order as collateral, it needs to be valued as the worse case scenario. Also we need to consider the possibility that the collateral position is not immediately removed after the fill or cancel.

We can only assume that the order will never exist in a state where underlying is less valuable than tickets. That price is explicitly disallowed by the program, and if the market moves in the direction, the order is guaranteed to fill before they are 1-to-1. Likewise, we know the quote quantity will always be less than the base quantity.

Imagine the price of tickets suddenly plummets and the order can never be filled, only cancelled. Now the user has a collateral position that is much more valuable than the tickets they locked up to sell it, and they can use this fake collateral to steal from lenders. So to mitigate this risk we should price the collateral as tickets, since we know that tickets will never be more valuable than the underlying.

Imagine the price of tickets is moving up. The order fills and the user receives underlying, and the price of tickets continues to go up. If the size of the collateral is counted in base quantity and valued as tickets, then the collateral value is at parity with the underlying received by the user at the moment of fill. But once the price of tickets goes up at all from the fill price, suddenly this collateral position is more valuable than the underlying they are going to receive. It may be in the user's best interest to prevent settlement in that situation, and use the collateral to steal from lenders in another market. The only way to guarantee this will be avoided is by giving a quantity of collateral tickets that is equal to quote quantity of the order.

There four options for how to assign value to collateral for a sell order:
1. priced as tickets, counted in base qty
2. priced as tickets, counted in quote qty
3. priced as underlying, counted in base qty
4. priced as underlying, counted in quote qty

#2 is the most conservative, and this conservative approach is mandated to guarantee that the aforementioned exploits are impossible.

programs/fixed-term/src/margin/instructions/initialize_margin_user.rs
zachanon: I wanted to deny access to modifying the `asset` and `debt` fields without going through the api in the `MarginUser`. When going through the code, it got a bit complex to know exactly what sort of operations you wanted to do per each instruction, and I decided a more instruction-specific api was a good way of encapsulating this. Do you have a good way to implement this change and keep the macro? If I set it to `pub(crate)` it would still be manipulable in the program itself.

programs/fixed-term/src/margin/instructions/initialize_margin_user.rs
zachanon: can you elaborate a bit on what you mean here?

programs/fixed-term/src/margin/instructions/margin_sell_tickets_order.rs
dnut: Yeah that sounds right to me.

For a borrow order, the collateral can be valued as the underlying because there is a claim that is larger in size with the same price, which fully offsets the collateral under any circumstances.

When selling tickets, there is no offsetting position. Instead the issued collateral token represents a user's deposited tickets, which has no guaranteed value relative to the underlying.

programs/fixed-term/src/margin/instructions/initialize_margin_user.rs
dnut: Gotcha that's a good point and these priorities probably override the minor advantages of using init. 

That being said, I think I see another way of accomplishing your goal. Not sure if it's actually better overall, but I just thought I'd mention it, since it does meet your priorities while also being compatible with init. I'm gonna take a closer look at the individual methods and see if I have any more thoughts.

You could limit the visibility of any methods and fields of the Debt and Asset structs that can be used to improperly mutate its state.

The downside is a slight compromise in separation of concerns. The Debt and Asset types would contain visibility restrictions that don't serve the type being restricted, but instead serve the safety of another type. Since we know they are only used in one place, and it's in the same file, it's not really a big deal, but it's still a little weird.

The upside is that you can safely expose the debt and asset fields publicly in the Margin user, which means:
- pro: You can use init.
- pro: It may be easier for clients to read the state of a MarginUser account.
- pro: You don't need a bunch of methods on MarginUser that do nothing except act as getters for data within the debt and asset fields, which feels a bit cluttered, and comes with a minor maintenance cost.
- con: MarginUser relies on separate type definitions to ensure its own safety  (complementary to previously mentioned downside)

programs/fixed-term/src/margin/instructions/margin_sell_tickets_order.rs
```diff
@@ -48,10 +48,14 @@ pub fn handler(ctx: Context<MarginSellTicketsOrder>, params: OrderParams) -> Res
             .map(|a| a.key()),
         CallbackFlags::MARGIN,
     )?;
+
+    // collateral accounting
+    let posted_ticket_value = order_summary.base_posted();
```
dnut, line 53: I see you changed this from quote_posted to base_posted since my last comment. I think you originally had it correct with quote_posted. It's the collateral mint that I was suggesting to change. The user should be issued ticket collateral, not underlying collateral.

The general idea is that the order might settle to either tickets or underlying. To be completely safe, it needs to be priced as the less valuable one (tickets) and counted as the less numerous one (underlying).

These are the four scenarios I mentioned in my previous post:

1. priced as tickets, counted in base qty - risky
2. priced as tickets, counted in quote qty - conservative
3. priced as underlying, counted in base qty - harmful
4. priced as underlying, counted in quote qty - risky

Originally you had this implemented as 4, which was risky because it has the *potential* to overvalue the position. You've changed it to 3, which is actually more dangerous, because the collateral position is *always* valued higher than anything the user could possibly get from the order. 2 is the only option that guarantees the collateral position is never overvalued.

dnut, line 53: I see you changed this from quote_posted to base_posted since my last comment. I think you originally had it correct with quote_posted. It's the collateral mint that I was suggesting to change. The user should be issued ticket collateral, not underlying collateral.

The general idea is that the order might settle to either tickets or underlying. To be completely safe, it needs to be priced as the less valuable one (tickets) and counted as the less numerous one (underlying).

These are the four scenarios I mentioned in my previous post:

1. priced as tickets, counted in base qty - risky
2. priced as tickets, counted in quote qty - conservative
3. priced as underlying, counted in base qty - harmful
4. priced as underlying, counted in quote qty - risky

Originally you had this implemented as 4, which was risky because it has the *potential* to overvalue the position. You've changed it to 3, which is actually more dangerous, because the collateral position is *always* valued higher than anything the user could possibly get from the order. 2 is the only option that guarantees the collateral position is never overvalued.

dnut: To elaborate on why it's always overvalued...

Let's say tickets are worth $0.90 and underlying is worth $1. The user places an order to sell 110 tickets for 100 USDC. The tickets they deposited are currently worth $99, and if the order matches, they will get $100 worth of underlying.

With collateralization approach 3, you issue them 110 tokens of underlying_collateral. Their collateral position is valued at $110 even though they could never get that something that valuable out of their order.

programs/fixed-term/src/margin/instructions/settle.rs
```diff
@@ -77,47 +84,68 @@ pub struct Settle<'info> {
...
-    let debt = ctx.accounts.margin_user.debt.total();
-    let ctokens_deserved = assets.collateral()?;
+    let debt = ctx.accounts.margin_user.total_debt();
 
-    // Notify margin of the current debt owed to fixed-term market
     if claim_balance > debt {
         ctx.burn_notes(
             &ctx.accounts.claims_mint,
-            &ctx.accounts.claims,
+            ctx.accounts.claims.to_account_info(),
```
dnut, line 94: It's redundant to call `to_account_info` here. It will be invoked by the MarketTokenManager. There are 6 similar redundant calls in this file.

programs/fixed-term/src/margin/instructions/settle.rs
```diff
@@ -77,47 +84,68 @@ pub struct Settle<'info> {
...
     }
     if ctokens_held < ctokens_deserved {
         ctx.mint(
-            &ctx.accounts.ticket_collateral_mint,
-            &ctx.accounts.ticket_collateral,
+            &ctx.accounts.underlying_collateral_mint,
+            ctx.accounts.underlying_collateral.to_account_info(),
             ctokens_deserved - ctokens_held,
         )?;
     }
```
dnut, line 144: This duplicated logic could be extracted into a function with two parameters.

programs/fixed-term/src/margin/instructions/settle.rs
```diff
@@ -77,47 +84,68 @@ pub struct Settle<'info> {
...
```
dnut, line 145: This same logic is now duplicated three times. It could be extracted into a function with three parameters.

programs/fixed-term/src/orderbook/state/mod.rs
```diff
@@ -596,59 +607,90 @@ impl SensibleOrderSummary {
         }
     }
 
-    // todo defensive rounding - depends on how this function is used
-    pub fn quote_posted(&self) -> Result<u64> {
-        fp32_mul(self.summary.total_base_qty_posted, self.limit_price)
-            .ok_or_else(|| error!(FixedTermErrorCode::FixedPointDivision))
+    /// Value of the posted portion of the order denominated in underlying (quote) tokens
+    pub fn quote_posted(&self, action: RoundingAction) -> Result<u64> {
```
dnut, line 611: I think it makes more sense to accept RoundingDirection as a parameter here rather than RoundingAction. It's easy to call `.direction()` at the call site. Requiring the order summary implementation to deal with unnecessary call-site context breaks the separation of concerns. There is more cognitive load when trying to understand the responsibility of the order summary.

I don't think the order summary needs to care about concepts like how to issue the various positions related to a loan, which is the information encoded in the conversion from RoundingAction to RoundingDirection. This method just needs to do some calculation about the quote quantity of the order. I would keep the calculation simple, just tell it which direction to round. The context can determine the necessary rounding direction and pass that along.

zachanon: Passing in the rounding action is the context. You only ever call this on specific actions, i.e. you had a borrow filled, you had a loan posted, etc. There is another function that takes rounding directions directly, but in all cases I have found it reduces cognitive burden to think about the action you are taking-- oh, I've just posted a borrow order -- rather than thinking about whether you want to shave a lamport, which is in fact determined by the action you are taking.

programs/fixed-term/src/orderbook/state/rounding.rs
```diff
@@ -0,0 +1,82 @@
+use std::convert::TryInto;
+
+use anchor_lang::{error, Result};
+
+use crate::FixedTermErrorCode;
+
+/// Describes the action being performed, in order to determine the rounding direction
+pub enum RoundingAction {
```
dnut, line 8: I'm struggling to understand how this type is helpful.

Each RoundingAction represents a particular context where a RoundingDirection needs to be instantiated. You need to understand the details of that particular context to determine what RoundingDirection you'll need in that context. RoundingAction doesn't remove that need for specificity. But it adds a layer of indirection, obscuring the relationship between a context and its particular rounding direction, without generalizing it. I think it would be clearer to just directly write the desired RoundingDirection in the context where it is needed.

The problem is amplified when it becomes unclear or impossible to select a RoundingDirection for a particular RoundingAction. One context may involve multiple calculations, and you may need different rounding directions for different calculations in that context. It really depends on the specifics of the context. To keep this pattern going, you need to add more RoundingAction variants with increasingly esoteric names.

programs/fixed-term/src/orderbook/state/mod.rs
dnut: Well, in this comment, I was not proposing anything other than changing the type of the parameter in this function. I was not saying anything about the call site. In the call site you could still do something like `RoundingAction::FillBorrow.direction()`. So you get the same exact benefit that you're already advocating for. It just makes this function definition less confusing.

However...

> I have found it reduces cognitive burden to think about the action you are taking-- oh, I've just posted a borrow order -- rather than thinking about whether you want to shave a lamport, which is in fact determined by the action you are taking.

I think this cognitive burden is a good thing because it encourages code correctness. The best way of rounding is extremely specific to the context. It's not safe to assume that the rounding direction you used for a prior calculation is also going to be applicable for a new calculation. Any time you add a new place where you're calculating the quote quantity, you should be forced to think about the rounding direction. 

I actually just posted another comment about this topic on the RoundingAction type.

programs/fixed-term/src/margin/instructions/settle.rs
zachanon: Not easily. Each case is requiring different mints and targets. The only way I can think of to abstract this involves a macro, and one that wouldn't exactly be easier to read than status quo. If you have a better way, I'm game.

programs/fixed-term/src/margin/instructions/settle.rs
zachanon: The compiler does not like it
```
error[E0277]: the trait bound `std::boxed::Box<anchor_lang::prelude::Account<'_, anchor_spl::token::TokenAccount>>: std::convert::AsRef<anchor_lang::prelude::AccountInfo<'_>>` is not satisfied
   --> programs/fixed-term/src/margin/instructions/settle.rs:141:13
    |
139 |         ctx.mint(
    |             ---- required by a bound introduced by this call
140 |             &ctx.accounts.underlying_collateral_mint,
141 |             ctx.accounts.underlying_collateral,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::AsRef<anchor_lang::prelude::AccountInfo<'_>>` is not implemented for `std::boxed::Box<anchor_lang::prelude::Account<'_, anchor_spl::token::TokenAccount>>`
    |
    = help: the trait `std::convert::AsRef<T>` is implemented for `std::boxed::Box<T, A>`
    = note: required for `std::boxed::Box<anchor_lang::prelude::Account<'_, anchor_spl::token::TokenAccount>>` to implement `anchor_lang::ToAccountInfo<'_>`
```

It has to be a box for stack size reasons, and it has to be a token account to grab the amount. You could use the accessor I guess, may as well implement that...

zachanon: I was wrong, cannot use the accessor, at least not easily. It requires a reference, which then causes the function to complain about mixing static and mutable references

programs/fixed-term/src/orderbook/state/rounding.rs
zachanon: > But it adds a layer of indirection, obscuring the relationship between a context and its particular rounding direction, without generalizing it. I think it would be clearer to just directly write the desired RoundingDirection in the context where it is needed.

I disagree. There are only so many contexts in which we are reading the quote from the book, its a rather low number. 

>One context may involve multiple calculations, and you may need different rounding directions for different calculations in that context. It really depends on the specifics of the context. To keep this pattern going, you need to add more RoundingAction variants with increasingly esoteric names.

Completely hypothetical and does not appear in the code. I cannot come up with a mundane case where we are reading the quote in different ways from one action. Instructions are fairly self contained, and you are only reading the summary when something gets posted, filled or cancelled.

programs/fixed-term/src/margin/instructions/initialize_margin_user.rs
zachanon: going to leave as is, this PR has sat for a while now

programs/fixed-term/src/margin/instructions/settle.rs
dnut: not important. up to you. but this works:
```rust
pub fn handler(ctx: Context<Settle>) -> Result<()> {
    // Notify margin of the current debt owed to fixed-term market
    set_note_balance(
        &ctx,
        &ctx.accounts.claims_mint,
        &*ctx.accounts.claims,
        ctx.accounts.margin_user.total_debt(),
    )?;

    // Notify margin of the amount of collateral that will in the custody of
    // tickets after this settlement
    set_note_balance(
        &ctx,
        &ctx.accounts.ticket_collateral_mint,
        &*ctx.accounts.ticket_collateral,
        ctx.accounts.margin_user.ticket_collateral()?,
    )?;

    // Notify margin of the amount of collateral that will in the custody of
    // tokens after this settlement
    set_note_balance(
        &ctx,
        &ctx.accounts.underlying_collateral_mint,
        &*ctx.accounts.underlying_collateral,
        ctx.accounts.margin_user.underlying_collateral(),
    )?;

    // Disburse entitled funds due to fills
    let entitled_tickets = ctx.accounts.margin_user.entitled_tickets();
    if entitled_tickets > 0 {
        verify_settlement_account_registration(
            &*ctx.accounts.margin_account.load()?,
            ctx.accounts.ticket_mint.key(),
            ctx.accounts.ticket_settlement.key(),
            FixedTermErrorCode::TicketSettlementAccountNotRegistered,
        )?;
        ctx.mint(
            &ctx.accounts.ticket_mint,
            &ctx.accounts.ticket_settlement,
            entitled_tickets,
        )?;
    }

    let entitled_tokens = ctx.accounts.margin_user.entitled_tokens();
    if entitled_tokens > 0 {
        verify_settlement_account_registration(
            &*ctx.accounts.margin_account.load()?,
            ctx.accounts.market.load()?.underlying_token_mint.key(),
            ctx.accounts.underlying_settlement.key(),
            FixedTermErrorCode::UnderlyingSettlementAccountNotRegistered,
        )?;
        ctx.withdraw(
            &ctx.accounts.underlying_token_vault,
            &ctx.accounts.underlying_settlement,
            entitled_tokens,
        )?;
    }

    // Update margin user assets to reflect the settlement
    ctx.accounts.margin_user.settlement_complete();
    ctx.accounts.margin_user.emit_all_balances()?;

    return_to_margin(
        &ctx.accounts.margin_account.to_account_info(),
        &AdapterResult {
            position_changes: vec![],
        },
    )
}

fn set_note_balance<'info>(
    ctx: &impl MarketTokenManager<'info>,
    mint: &UncheckedAccount<'info>,
    token_account: &Account<'info, TokenAccount>,
    target_amount: u64,
) -> Result<()> {
    let current_balance = token_account.amount;
    if current_balance > target_amount {
        ctx.burn_notes(mint, token_account, current_balance - target_amount)?;
    }
    if current_balance < target_amount {
        ctx.mint(mint, token_account, target_amount - current_balance)?;
    }

    Ok(())
}
```
Actually I think this abstraction makes sense as part of MarketTokenManager. In that case I would make the parameters more generic, maybe something like this:
```rust
    fn set_note_balance(
        &self,
        mint: impl ToAccountInfo<'info> + Copy,
        token_account: impl ToAccountInfo<'info> + Copy,
        target_amount: u64,
    ) -> Result<()> {
        let current_balance = token::accessor::amount(&token_account.to_account_info())?;
        if current_balance > target_amount {
            self.burn_notes(mint, token_account, current_balance - target_amount)?;
        }
        if current_balance < target_amount {
            self.mint(mint, token_account, target_amount - current_balance)?;
        }

        Ok(())
    }
```
and the calls would look almost the same
```rust
    // Notify margin of the current debt owed to fixed-term market
    ctx.set_note_balance(
        &ctx.accounts.claims_mint,
        &*ctx.accounts.claims,
        ctx.accounts.margin_user.total_debt(),
    )?;

    // Notify margin of the amount of collateral that will in the custody of
    // tickets after this settlement
    ctx.set_note_balance(
        &ctx.accounts.ticket_collateral_mint,
        &*ctx.accounts.ticket_collateral,
        ctx.accounts.margin_user.ticket_collateral()?,
    )?;

    // Notify margin of the amount of collateral that will in the custody of
    // tokens after this settlement
    ctx.set_note_balance(
        &ctx.accounts.underlying_collateral_mint,
        &*ctx.accounts.underlying_collateral,
        ctx.accounts.margin_user.underlying_collateral(),
    )?;
```

programs/fixed-term/src/margin/instructions/settle.rs
dnut: fyi `&*` converts a Box to a reference

programs/fixed-term/src/orderbook/instructions/consume_events/handler.rs
```diff
@@ -149,12 +154,13 @@ fn handle_margin_fill<'info>(
...
-                user.assets.entitled_tokens.try_add_assign(disburse)?;
-                let strike_timestamp = Clock::get()?.unix_timestamp;
-                let maturation_timestamp = strike_timestamp.safe_add(tenor as i64)?;
-                let sequence_number = user
-                    .debt
-                    .new_term_loan_from_fill(base_size, maturation_timestamp)?;
+                let sequence_number = user.maker_fill_borrow_order(
+                    true,
+                    disburse,
+                    quote_size,
```
dnut, line 160: Relating to my long winded other comment about RoundingAction, this is where I think we actually need to input a rounded *up* value, despite needing to use this as a rounded *down* value up above when deciding how much to disburse.

dnut: One approach would be to pass around the quote quantity as an fp32 and only round it at the last possible second when it's being accumulated into the posted_quote quantity or used for a token transfer.

programs/fixed-term/src/orderbook/state/rounding.rs
suscd: discussed offline, doesn't appear to be currently exploitable

suscd (approved):

PR #640, opened 2023-03-16
misc improvements + liquidator support

dnut:
fixed term program:
- Invoke the clock syscall earlier in the instruction and pass it down by value, rather than depending on syscalls randomly throughout the codebase. This makes the code easier to call outside the solana runtime, which is useful for tests and the liquidator.

margin program:
- Return invalid price instead of error when price is zero. This behavior was inconsistent and leads to premature errors. Invalid prices are already handled properly later on when they are used.
- During refresh_deposit_position, optionally refresh the balance. A deposit position is a normal ATA, so it should be customary to deposit tokens with a standard SPL token transfer from any other wallet. It's also possible that the settle instruction in Fixed Term could update the balance without refreshing the position. It's unintuitive that refreshing the position would not refresh its balance to the latest value. Now you have to explicitly say "false" to avoid refreshing the balance. All the library functions that execute a single refresh expose the boolean. The library functions that execute refreshes on all positions set it to true.

simulation:
- RpcConnection: reorganize struct internals to make them easier to manipulate. All the fields remain private.
- Add clone_with_payer to SolanaRpcClient so you can use the client with different payers

margin-sdk + instructions:
- new and improved PositionRefresher
  - more generic - can be used to represent an individual margin account, or to refresh any margin account by address or by its already fetched state.
  - SmartRefresher enables composition of multiple refreshers:
    - more intuitively, by encapsulating delegation logic into a single place
    - with better performance, since it can query for the account only once and pass along the state to other refreshers
  - A canonical position refresher is defined clearly with a function that only needs an rpc client, so any code can easily refresh any positions with the minimum dependencies.
- add MarginInvokeContext as a simpler alternative to MarginTxBuilder for adapter invocation, with less dependencies.
- move some sdk method logic into delegated functions so they are usable without needing a big builder struct

environment:
- Builder
  - now uses an enum to represent whether the proposed instructions will be executed directly or through a proposal. previously these were just redundant separate fields that could get out of sync, now it's only one or the other. 
  - it is now impossible to construct a Builder without the correct authority/proposal context.
  - the correct authority is used consistently. The proposal_authority method had a bug which is now corrected, to use the authority instead of the payer.
- make the code more modular
- make some functions public so tests can use them directly, since they already contain test-specific logic.

hosted:
- use different addresses for payer, crank, and airspace authority in test contexts to improve test coverage
- Consolidate most of TestContext into MarginTestContext, so a common approach is used for the basic airspace initialization. TestContext remains as a thin wrapper around MarginTestContext to initialize additional state and provide a client instance. This improves interoperability between different testing approaches, and sets the stage to consolidate other test code to a more unified approach.
- split MarginTestContext helper methods to separate files


dnut:
notes from discussion
- token account should be optional for refresh deposit position - and this may fix e2e tests
- constructors can be used to improve the builder authority situation. ensure the field, method, and proposal context are consistent with each other, and fully instantiated properly the moment the type is constructed

suscd (approved):

PR #642, opened 2023-03-17
Allow a service to 'auto roll' a `TermLoan`

zachanon:
Creates a permissionless instruction that allows a user's debt to be serviced by attempting to borrow from the market in order to pay down the outstanding balance on the `TermLoan`.


programs/fixed-term/src/margin/state.rs
```diff
@@ -544,7 +548,18 @@ impl Default for Assets {
...
 #[derive(Zeroable, Default, Debug, Clone, PartialEq, Eq, AnchorSerialize, AnchorDeserialize)]
-pub struct AutoRollConfig {
+pub struct BorrowAutoRollConfig {
+    /// the limit price at which orders may be placed by an authority
+    pub limit_price: u64,
+
+    /// The 'auto-roll' function for borrowers is determined by a user-controlled tenor. The auto-roll
+    /// service will attempt to auto-roll any orders with a matured `roll_tenor`, regardless of the
+    /// maturity of the `TermLoan`
+    pub roll_tenor: u64,
```
dnut, line 558: This seems complicated. The margin account needs a MarginUser in a market with the target tenor. The auto-roll instruction needs to work with two Markets, two MarginUsers, etc. But I don't see any of that in the program. This field doesn't serve any purpose right now so why is it included?

programs/fixed-term/src/margin/instructions/auto_roll_borrow_order.rs
```diff
@@ -0,0 +1,157 @@
...
+    #[account(mut)]
+    pub claims: AccountInfo<'info>,
+
+    /// Token mint used by the margin program to track the debt that must be collateralized
+    #[account(mut, address = orderbook_mut.claims_mint() @ FixedTermErrorCode::WrongClaimMint)]
+    pub claims_mint: AccountInfo<'info>,
+
+    /// Token account used by the margin program to track the debt that must be collateralized
+    #[account(mut)]
+    pub underlying_collateral: AccountInfo<'info>,
```
dnut, line 37: I think this needs a has_one constraint on the margin_user

programs/fixed-term/src/margin/instructions/auto_roll_borrow_order.rs
```diff
@@ -0,0 +1,157 @@
...
+            term_loan: &self.new_loan,
+            margin_account: &self.margin_account.to_account_info(),
+            claims: &self.claims,
+            claims_mint: &self.claims_mint,
+            underlying_collateral: &self.underlying_collateral,
+            underlying_collateral_mint: &self.underlying_collateral_mint,
+            underlying_token_vault: &self.underlying_token_vault,
+            fee_vault: &self.fee_vault,
+            // returns funds directly to the market
+            underlying_settlement: &self.underlying_token_vault,
```
dnut, line 98: So the program is going to do a token transfer from the vault to the vault? The repayment works different, with `internal_call` true, it assumes that the repayment is already in the vault, and bypasses the token transfer. Personally I would have probably chosen one of those two approaches and applied it to to both steps so it's easier to reason about the behavior of the instruction overall.


With the bypass approach, I would probably use parameter names that are very clear about their effect:
- `repayment_is_already_in_vault` and `loan_is_already_disbursed`
- `skip_repay_transfer` and `skip_disburse_transfer`

Actually ideally I think the funds would get pushed through some other token account that doesn't have funds owned by other users. This would make the code more robust to vulnerabilities being introduced. But it requires us to come up with another token account which is complexity we may not need.

Anyway the code seems fine. This is just a stylistic point and I imagine there are endless arguments in every direction. Just wanted to share my perspective.

programs/fixed-term/src/margin/state.rs
zachanon: For auto rolling a loan, we run into a problem where we race against the liquidator. If the loan is past due, we need to let the liquidator do its job.

So, one way to solve this is a user configured tenor to attempt to roll at. The servicer will see that the roll tenor has elapsed and roll the loan, before it reaches maturity and needs liquidation. 

What do you mean by "two Markets, two MarginUsers"?

dnut: I completely misunderstood what "roll_tenor" means. I understood this as a request to have the loan repaid by rolling it into loan from a market with a different tenor.

Is the roll tenor being enforced anywhere? We should verify that a loan is not auto-rolled until after the roll tenor has been reached.

suscd: now enforced

PR #657, opened 2023-04-05
set common rust dependencies in workspace

suscd:


suscd:
sbf build is only on rustc 1.62

suscd:
> @suscd
> 
> Is this a new feature in cargo? I looked into this a while back and concluded it was impossible.
> 
> Why did you close it?

the default toolchain for solana programs uses cargo version 1.62, which doesn't support this feature (available in 1.64)

PR #661, opened 2023-04-07
adds backpack wallet

qiqicodes:
adds wallet and logo


apps/react-app/src/App.tsx
```diff
@@ -30,6 +31,7 @@ export const App = (): JSX.Element => {
   const wallets = useMemo(
     () => [
       // new PhantomWalletAdapter(),
+      new BackpackWalletAdapter(),
```
suscd, line 34: isn't this unnecessary?

qiqicodes: it is not necessary, the adapter is placed here for user with no backpack wallet to see this wallet as an option.

davodesign84:
It seems to me that the wallet is actually working correctly on master. The only issue is the lack of icon which is added on this PR. This line is also superfluous https://github.com/jet-lab/jet-v2/pull/661/files#diff-24ed2e29abdb01854048bc8e18a9f42cce3693fe09371eb4716fb517e62a938cR7 as the adapter is imported but not used aywhere.

davodesign84 (approved):

davodesign84 (approved):

PR #676, opened 2023-04-19
logging improvements

dnut:
- debug_msg macro for programs to log in tests when feature is enabled (currently only used in fixed term repay)
- more granular logging targets in simulation to enable filtering

codecov must have a bug, its results are incorrect

### DIFF
--- a/libraries/rust/simulation/src/runtime.rs
+++ b/libraries/rust/simulation/src/runtime.rs
@@ -58,7 +58,7 @@ thread_local! {
 }
 
 declare_logging! {
-    logging = "sim" {
+    logging = "jet_simulation" {
         program         = "program";
         account_loader  = "acctldr";
         account_realloc = "realloc";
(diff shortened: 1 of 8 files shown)


libraries/rust/simulation/src/runtime.rs
```diff
@@ -57,6 +57,23 @@ thread_local! {
     static LOCAL_CONTEXTS: RefCell<Vec<LocalContext>> = RefCell::new(Vec::new());
 }
 
+declare_logging! {
+    logging = "sim" {
```
suscd, line 61: changing the top level logging target is kind of annoying, because it's now this extra bit of information you have to know and is difficult to infer. This should just be `jet_simulation`, which is more obvious

suscd (changes requested):

libraries/rust/simulation/src/runtime.rs
dnut: done

suscd (approved):

PR #678, opened 2023-04-20
Localnet script consolidation

dnut:
Consolidate `localnet_lib.sh` and `on_localnet.sh` into a single file called `on_localnet.sh`

The diff for this PR looks like I deleted a file with a lot of text, and added a lot of text to another file. But if you look at the individual commits you'll see that I renamed the file and barely changed any contents.

## Purpose

Originally I made the files separate because I wanted a shell script with a bunch of functions in it that could be used as either:
- a library, by sourcing the functions into another script or an interactive shell
- an executable, by passing a function name as an argument and having it run that function

The only way I could think to do that was with two separate files. But this indirection is cluttered and confusing. I figured out a way to make the single file serve both purposes.


qiqicodes (approved):
lgtm =]

suscd (approved):

PR #685, opened 2023-04-27
remove redundant logs from test runtime that cause segfault

dnut:
These logs seem redundant with the logging that happens right before. Maybe I'm missing something but it seems pointless. (update: I was missing something, see discussion below)

In two of the tests in the liquidator, this logging code results in a segfault. One test has the error at InvokeContext::get_log_collector and the other has it in the last line of LogCollector::log when it pushes to messages.

If there's really no point in these log messages, then I say we remove it, to reduce our reliance on unsafe code that has proven to be incorrect. Otherwise we need to figure out how the code is incorrect and fix it.

https://github.com/jet-lab/cranks/actions/runs/4811891241/jobs/8566453695?pr=49


suscd:
not sure about this, because the purpose of the log collector is to associate the logs with the particular transaction being executed, such that the transaction error returned includes the program logs (regardless of whether or not the rust logging facility is configured). By not using the collector, the only way to know why a transaction failed is to enable the rust logs from the simulation library. Maybe that's not such a big deal, but it does seem somewhat inconvenient. 

I've been mucking around in the simulator recently, will see if I can determine what the underlying issue is causing the crash

dnut:
OK, then this change is not ideal. I'll keep using this branch to test the liquidator though, since it's the only way I have right now to get all the tests to complete without error.

PR #686, opened 2023-04-27
Add autoroll field to events

davodesign84:
Depends on: https://github.com/jet-lab/jet-v2/pull/683


davodesign84:
> 

So apparently the events already include the flags, one of which is the autoroll flag. I think it might be better to parse that in jet-flight-logs instead of emitting redundant in the logs. 

Will look into the global config events.

suscd (approved):

PR #691, opened 2023-05-01
Setup and override the lookup authority on localnet

nevi-me:
- Add a lookup authority keypair as part of the localnet setup.
- Create and populate a registry for the authority
- Update the localnet config to use this authority

Tested locally by starting up the validator and confirming that:
- the config is updated
- the lookup registry and 2 lookup tables are created

### DIFF
--- a/libraries/rust/environment/src/client_config.rs
+++ b/libraries/rust/environment/src/client_config.rs
@@ -55,7 +55,7 @@ impl JetAppConfig {
     pub async fn from_env_config<I: NetworkUserInterface>(
         env: EnvironmentConfig,
         network: &I,
-        default_lookup_authority: Option<Pubkey>,
+        override_lookup_authority: Option<Pubkey>,
     ) -> Result<Self, ConfigError<I>> {
         let mut seen = HashSet::new();
         let mut tokens = vec![];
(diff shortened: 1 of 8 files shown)


libraries/rust/environment/src/client_config.rs
```diff
@@ -55,6 +55,7 @@ impl JetAppConfig {
     pub async fn from_env_config<I: NetworkUserInterface>(
         env: EnvironmentConfig,
         network: &I,
+        default_lookup_authority: Option<Pubkey>,
```
suscd, line 58: I think you should replace `default` with `override`, e.g. this parameter should be `override_lookup_authority`, which more accurately describes how its being used.

tools/ctl/src/lib.rs
```diff
@@ -284,6 +285,13 @@ pub enum Command {
         /// The path to write the generated file to
         #[clap(long, short = 'o')]
         output: PathBuf,
+
+        /// The address to use as a default lookup registry, obtained from a keypair
+        #[clap(long)]
+        default_lookup_keypair: Option<PathBuf>,
+        /// The address to use as a default lookup registry, supplied directly
+        #[clap(long)]
+        default_lookup_address: Option<Pubkey>,
```
suscd, line 294: seems simpler to just only take the pubkey, and leave reading the address for a key as the caller's problem (which isn't difficult, you can do it with `solana-keygen pubkey` in bash

PR #694, opened 2023-05-04
compliance related ui-tweaks

davodesign84:


davodesign84:
eheh good spot, I just copy / pasted the original text from the task 😬

PR #695, opened 2023-05-04
Swap/openbook referral fees + test service

nevi-me:


programs/margin-swap/src/instructions/openbook_swap.rs
```diff
@@ -62,6 +63,10 @@ pub struct OpenbookSwapInfo<'info> {
     /// CHECK:
     pub vault_signer: AccountInfo<'info>,
 
+    /// The referrer account owned by the control program
+    #[account(mut, token::authority = CONTROL_AUTHORITY)]
```
dnut, line 67: Isn't the control program deprecated? I thought airspaces supersede everything in there. Why can't the airspace authority or governor own this?

Also this is missing a constraint. I would validate that it is the specific account you expect to be used here, not just the authority.

programs/margin-swap/src/instructions/route_swap.rs
```diff
@@ -557,7 +557,7 @@ fn exec_swap_split<'info>(
             dst_ata.to_account_info()
         }
         SwapRouteIdentifier::OpenBook => {
-            let accounts = remaining_accounts.take(11).cloned().collect::<Vec<_>>();
+            let accounts = remaining_accounts.take(12).cloned().collect::<Vec<_>>();
```
dnut, line 560: It would be nice if we didn't have to manually maintain this number, but I'm not sure how to avoid it.

dnut (changes requested):

suscd (dismissed):

programs/margin-swap/src/instructions/openbook_swap.rs
suscd: Sort of, but it currently owns all the token accounts any fees get collected into, so for consistency it makes sense to just also collect these there. It probably makes sense to repurpose the control program to only handle fee processing in the future.

Though I agree that this should probably check that this is the correct ATA address instead of just the owner/authority.

dnut: It's easy to change the owner of a token account.

suscd: The check we were referring to is an address check to validate that it's an ATA, which ensures there's only a single account that fees can be sent to

suscd (approved):

PR #700, opened 2023-05-05
Refactor `AutoRollConfig` and add price bounds to `FixedTermMarket` orders

zachanon:
Adjusting the auto roll config no longer needs to run through an adaptor

Fixed term market orders have configurable bounds on limit price to prevent exploit and potential user error

### DIFF
--- a/programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
+++ b/programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
@@ -1,5 +1,6 @@
 use anchor_lang::prelude::*;
 use anchor_spl::token::{Token, TokenAccount};
+use jet_airspace::state::AirspacePermit;
 use jet_program_proc_macros::MarketTokenManager;
 
 use crate::{
@@ -10,6 +11,12 @@ use crate::{
 
 #[derive(Accounts, MarketTokenManager)]
 pub struct RedeemDeposit<'info> {
+    /// Metadata permit allowing this user to interact with this market
+    #[account(
+        constraint = permit.airspace == market.load()?.airspace @ FixedTermErrorCode::WrongAirspaceAuthorization,
+    )]
+    pub permit: Account<'info, AirspacePermit>,
+
     /// The tracking account for the deposit
     #[account(mut,
               close = payer,
--- a/programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
+++ b/programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
@@ -13,6 +13,7 @@ use crate::{
 pub struct RedeemDeposit<'info> {
     /// Metadata permit allowing this user to interact with this market
     #[account(
+        has_one = owner @ FixedTermErrorCode::WrongAirspaceAuthorization,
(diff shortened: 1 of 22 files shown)


suscd:
> Fixed term market orders have configurable bounds on limit price to prevent exploit and potential user error

Why? If the intent is to prevent user error, isn't that a problem for the UI? Why does the program need to care about what price a user sets?

zachanon:
> > Fixed term market orders have configurable bounds on limit price to prevent exploit and potential user error
> 
> Why? If the intent is to prevent user error, isn't that a problem for the UI? Why does the program need to care about what price a user sets?

This is a requested feature from @eyevz and I believe from ottersec as well. Limits the surface area for exploits by not being able to make negative interest rate trades, for example.

suscd:
> This is a requested feature from @eyevz and I believe from ottersec as well. Limits the surface area for exploits by not being able to make negative interest rate trades, for example.

How is this reducing the surface area for exploits?

Also even if you wanted to prevent using nonsensical prices, I don't understand why it needs to be configurable, isn't the price range always going to be `0.0 - 1.0` ?

zachanon:
> > This is a requested feature from @eyevz and I believe from ottersec as well. Limits the surface area for exploits by not being able to make negative interest rate trades, for example.
> 
> How is this reducing the surface area for exploits?
> 
> Also even if you wanted to prevent using nonsensical prices, I don't understand why it needs to be configurable, isn't the price range always going to be `0.0 - 1.0` ?

You cannot make an order with negative interest, for one. The idea being I suppose that if some vulnerability were to be found elsewhere, you may not be able to exploit it because the limit price range is bounded to sane values. I just wrote the current values as baselines to keep everything running, ideally we would set the bounds to be something more in line with the market.

I could see a case to be made for simply making the price range globally as between 0 and 1, but I implemented this as configurable here due to aforementioned request.

dnut:
I think it makes sense to hardcode (0,1] as the legal price range in the program. Actually I thought this was already implemented. This is sufficient for the protocol to protect itself. Hardcoding the value instead of exposing another configurable property keeps the program simple. Simplicity is valuable for maintenance, bug prevention, and reasoning about behavior.

Clients can protect users from their own mistakes. I don't see a need for the program to do that. Putting this constraint in the client makes it easier to change the configuration, instead of having to go through governance.

suscd:
some instructions still missing airspace checks:
* `exchange_tokens`
* `redeem_deposit`
* `stake_tickets`
* `sell_tickets_order`
* `initialize_margin_user`

programs/fixed-term/src/tickets/instructions/redeem_deposit.rs
```diff
@@ -10,6 +11,12 @@ use crate::{
 
 #[derive(Accounts, MarketTokenManager)]
 pub struct RedeemDeposit<'info> {
+    /// Metadata permit allowing this user to interact with this market
+    #[account(
+        constraint = permit.airspace == market.load()?.airspace @ FixedTermErrorCode::WrongAirspaceAuthorization,
+    )]
+    pub permit: Account<'info, AirspacePermit>,
```
suscd, line 18: missing a `has_one = owner` check

PR #715, opened 2023-05-16
better rpc error, test configuration

dnut:
- When a transaction fails in SolanaRpcClient::send_and_confirm, include the transaction signature in the anyhow error message. This has been critical for me to diagnose transaction failures in the liquidator.
- Allow the airspace to be customized for manual testing of the liquidator, so I don't need to configure the liquidator to use weird esoteric airspace names that differ for each test

the codecov result makes no sense because the code is used in a lot of tests


dnut:
the codecov result makes no sense because the code is used in a lot of tests

suscd (approved):

PR #729, opened 2023-05-26
make SplSwapPool::get_pools Send

dnut:
This is necessary for the liquidator to be able to stay up to date with the latest state. See [this change](https://github.com/jet-lab/cranks/pull/49/commits/2077a8db4e66a51a1731da1a2dc1b47f299ca019#diff-4f5c05f7e86915cdfb68b8c84c671eed9863d43de9dcb6cb6b985a5248617ea6R22-R31) - the trait itself must be Send, and the trait cannot be implemented with get_pools unless get_pools is Send.

I made the async function's Future implement Send when previously it did not. This adds a redundant deserialization of the SwapV1 struct, which may have some performance cost, but to me it looks negligible. Basically it's just a copy of 300 bytes plus some minimal branching.

### DIFF
--- a/libraries/rust/margin/src/swap/spl_swap.rs
+++ b/libraries/rust/margin/src/swap/spl_swap.rs
@@ -77,28 +77,26 @@ impl SplSwapPool {
 
         let mut pool_sizes = HashMap::with_capacity(accounts.len());
         for (swap_address, pool_account) in accounts {
-            let swap = SwapV1::unpack(&pool_account.data[1..]);
-            let swap = match swap {
-                Ok(swap) => swap,
-                Err(_) => continue,
+            let pool_mint = {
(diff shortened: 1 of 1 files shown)


libraries/rust/margin/src/swap/spl_swap.rs
```diff
@@ -77,28 +77,26 @@ impl SplSwapPool {
...
+                    .get(&swap.token_a_mint)
+                    .and_then(|_| supported_mints.get(&swap.token_b_mint))
+                    .is_none()
+                {
+                    continue;
+                }
+                swap.pool_mint
             };
-
-            if !swap.is_initialized {
```
dnut, line 86: I removed this because it is redundant with unpack, which returns an error if the account is not initialized.

PR #774, opened 2023-06-14
removing bulletted list of old features from disclaimer modal

KevinRafaelConnor:
- removes the list of outdated features from the disclaimer modal
- removes associated styling that is no longer necessary


KevinRafaelConnor:
@davodesign84 here is a preview of what it looks like after this change. After playing with the styling I thought it looked better with justified text so that the right side of the disclaimer text is not strangely visually unaligned with the left side. 

Pinging you to let me know if you prefer it with yet-different styling, as well as any other comments. Thank you.

without text justification: image
with justification: 
image

